PR #1, opened 2021-08-05
Support multiple packages in one entry.

oliverchang:
- Rename "affects" to "affected", and make it a list of objects.
- Move "package", "ecosystem_specific", "database_specific" to "affected".
- Add "routines" and "platforms" to "affected" as well.

### DIFF
--- a/schema.md
+++ b/schema.md
@@ -1,6 +1,6 @@
 # Open Source Vulnerability format
 
-**Version 0.7.1 (June 30, 2021)**
+**Version 0.8 (August 5, 2021)**
 
 Original authors:
 - Oliver Chang (ochang@google.com)
@@ -56,28 +56,30 @@ contain UTF-8 text.
 	"withdrawn": string,
 	"aliases": [ string ],
 	"related": [ string ],
-	"package": {
-		"ecosystem": string,
-		"name": string,
-		"purl": string,
-	},
 	"summary": string,
 	"details": string,
-	"affects": {
+	"affected": [ {
+		"package": {
+			"ecosystem": string,
+			"name": string,
+			"purl": string,
+		},
 		"ranges": [ {
 			"type": string,
 			"repo": string,
 			"introduced": string,
 			"fixed": string
 		} ],
-		"versions": [ string ]
-	},
+		"versions": [ string ],
+		"platforms": [ string ],
+		"routines": [ string ],
+		"ecosystem_specific": { see description },
+		"database_specific": { see description },
+	} ],
 	"references": [ {
 		"type": string,
 		"url": string
 	} ],
-	"ecosystem_specific": { see description },
-	"database_specific": { see description },
 }
 ```
 
@@ -158,11 +160,59 @@ same database, the duplicate entry could be written using only the `id`,
 The `related` field gives a list of IDs of closely related vulnerabilities, such
 as the same problem in alternate ecosystems. 
 
-### package field
+### summary, details fields
 
-The `package` field is a JSON object identifying the affected code library or
-command provided by the package. The object itself has two required fields,
-`ecosystem` and `name`, and an optional `purl` field.
+The `summary` field gives a one-line, English textual summary of the
+vulnerability. It is recommended that this field be kept short, on the order of
+no more than 120 characters.
+
+The `details` field gives additional English textual details about the
+vulnerability. The field is plain text. Newline characters must be considered
+line breaks when displaying the text, but the display need not use a fixed-width
+font, and the text should not assume one.
+
+The `summary` field is plain text.
+
+The `details` field is CommonMark markdown (a subset of GitHub-Flavored
+Markdown). Display code may at its discretion sanitize the input further, such
+as stripping raw HTML and links that do not start with http:// or https://.
+Databases are encouraged not to include those in the first place. (The goal is
+to balance flexibility of presentation with not exposing vulnerability database
+display sites to unnecessary vulnerabilities.)
+
+### affected fields
+
+The `affected` field is a JSON array containing objects that describes the
+affected packages versions, meaning those that contain the vulnerability.
+
+Within each object in the `affected` array, the `package` field identifies the
+package containing the vulnerability.
+
+The `versions` field can enumerate a specific set of affected versions, and the
+`ranges` field can list ranges of affected versions, under a given defined
+ordering. A version is considered affected if it lies within any one of the
+ranges or is listed in the versions list.
+
+The `versions` list should - with one exception - always be present, to allow
+software to answer the question "is this specific version affected?" without
+having to contain code specific to every different ecosystem. The one exception
+is if the affected versions are valid SemVer 2.0 versions which can be
+accurately summarized by one or more non-overlapping SemVer ranges. In that
+case, the SemVer ranges can be listed instead, in entries in the `ranges` field
+with type `SEMVER` (see below). In this case, the SemVer ranges act as a kind of
+compact form of a larger `versions` list. Ecosystems that do not use SemVer
+identifiers or that order versions differently from SemVer must include the
+enumerated `versions` list, although they can also add ranges of type
+`ECOSYSTEM` for additional context.
+
+In short, each object in the `affected` array must contain either a non-empty
+`versions` list or at least one range in the `ranges` list of type `SEMVER`.
+
+#### affected[].package field
+
+The `affected` object's `package` field is a JSON object identifying the
+affected code library or command provided by the package. The object itself has
+two required fields, `ecosystem` and `name`, and an optional `purl` field.
 
 The `ecosystem` identifies the overall library ecosystem. It must be one of the
 strings in the table below. The `name` field is a string identifying the library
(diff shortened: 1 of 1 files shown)


schema.md
```diff
@@ -158,11 +158,64 @@ same database, the duplicate entry could be written using only the `id`,
...
+Markdown). Display code may at its discretion sanitize the input further, such
+as stripping raw HTML and links that do not start with http:// or https://.
+Databases are encouraged not to include those in the first place. (The goal is
+to balance flexibility of presentation with not exposing vulnerability database
+display sites to unnecessary vulnerabilities.)
+
+### affected fields
+
+The `affected` field is a JSON array containing objects that describes the
+affected packages versions, meaning those that contain the vulnerability.
```
rsc, line 184: s/packages/package/

schema.md
```diff
@@ -158,11 +158,64 @@ same database, the duplicate entry could be written using only the `id`,
...
+
+Within each object in the `affected` array, the `package` field identifies the
+package containing the vulnerability. In most cases, there should be exactly one
+entry in the `affected` array per affected `package` to describe all affected
+versions. In rare cases, for example if the `ecosystem_specific` encodes
+platform information that doesn't apply equally to all listed versions and
+ranges, a separate entry with the same `package` in the `affected` array may be
+used.
```
rsc, line 192: s/used/needed/

schema.md
```diff
@@ -158,11 +158,64 @@ same database, the duplicate entry could be written using only the `id`,
...
+
+The `versions` field can enumerate a specific set of affected versions, and the
```
rsc, line 194: I wonder if we should collapse ranges and versions down to

   "versions": [{"type":..., "version":..., "before":...}] 

like in https://github.com/CVEProject/cve-schema/issues/87#issuecomment-893978107, with `"type": "exact"` for a single-version range, both to match that and to avoid having the two different fields that have to be merged by implementations.

On the other hand then you'd have to say that it's OK to have "exact" versions overlap non-semver ranges, to make them available to tooling.

schema.md
```diff
@@ -158,11 +158,64 @@ same database, the duplicate entry could be written using only the `id`,
...
+`ranges` field can list ranges of affected versions, under a given defined
+ordering. A version is considered affected if it lies within any one of the
+ranges or is listed in the versions list.
+
+The `versions` list should - with one exception - always be present, to allow
```
rsc, line 199: Still, this discussion would get a lot simpler. No exceptions.

schema.md
```diff
@@ -762,3 +769,23 @@ Based on that, added some new reference types.
...
+where multiple packages are supported in a single entry. We've
+also been suggesting changes to the CVE schema  for better alignment
+([1](https://github.com/CVEProject/cve-schema/issues/86),
+ [2](https://github.com/CVEProject/cve-schema/issues/87),
+ [3](https://github.com/CVEProject/cve-schema/issues/88),
+ [4](https://github.com/CVEProject/cve-schema/issues/89)).
+
+This is a breaking change, but we hope to make migration easier by renaming the
+"affects" field to "affected" to allow existing consumers and producers of this
+data to more easily handle old and new versions of entries.
```
rsc, line 791: It may be worth a comment like "We are still gathering experience about this new arrangement, and it may change again in the near future."

schema.md
oliverchang: Would prefer to keep the current explicit separation for our schema. Enforcing the that "versions" has to always exist if not SEMVER is much easier this way (as you mention below).

It should be pretty trivial for tooling to convert between the two, and too much on top of handling the other differences.

schema.md
oliverchang: done!

schema.md
oliverchang: done!

schema.md
oliverchang: ack.

schema.md
oliverchang: Sure, added. But I'd rather we avoid future breaking changes unless absolutely necessary from this point. We're at a point with many consumers that expect stability and every breaking change (especially one as substantial as this) is quite painful.

rsc: Then perhaps we should wait a bit on committing this until we get more experience from Go and OSI or anyone else?

inferno-chromium: yes i don't think we should even make this change before consulting with the current consumers like Go, OSI, Rust. Can we get their opinion. is this a big usecase ?

inferno-chromium (approved):

oliverchang:
Also moved `repo` as discussed: https://github.com/ossf/osv-schema/pull/1/commits/67693a7111bd4813a5a1d79eb63a4ec3d87ed831

joshbressers:
@oliverchang To followup from our conversation yesterday (or probably your today)

I mentioned I didn't like the new "affected" and I realize why now

```
	"affected": [ {
		"package": {
			"ecosystem": string,
			"name": string,
			"purl": string,
		},
		"ranges": [ {
			"type": string,
			"repo": string,
			"events": [ {
				"introduced": string,
				"fixed": string
			} ]
		} ],
		"versions": [ string ],
		"ecosystem_specific": { see description },
		"database_specific": { see description },
	} ],
```

I think the way CVE overloads the package field in an ID is wrong in the modern world. It made sense 20 years ago

Here is the example that made me realize this. It's the same problem for copy and pasting code, or embedding source files, so I ask whoever reads this not to nitpick it too much. It's a hard problem.

Let's say I have a npm project, npm will embed the dependencies, so for example it's VERY common for a project to ship multiple copies of the lodash library as different packages can depend on different versions.

If there is one vulnerability in lodash, today it would get one ID and you would see your application is affected by the same ID multiple times. This means you actually have to key off package:ID, because just the ID isn't enough useful information.

Now, package:ID as the key may not be the worst thing in the world, but it probably creates some new challenges we don't understand. Today I think just using an ID is the most common use case. I want to think about this more, but I wanted to suggest that allowing an array of packages should not be taken lightly.

kaniini:
I don't follow this argument, any scanner worth anything should automatically dedup any reported vulnerabilities, I think.

joshbressers:
> I don't follow this argument, any scanner worth anything should automatically dedup any reported vulnerabilities, I think.

It's not about the dupes so much as it is about having a way to understand what is vulnerable to a given problem. I could be off my rocker here as well, this makes sense in my head.

Let me just make something up to illustrate my thinking. This is a little unrealistic I know.

Let's say I have a project, and there are two different libraries that run as webservers to service REST requests. The two libraries each share one common source file that is vulnerable to a RCE. In today's model, that would be one ID both libraries share, even though they are unrelated.

Is that acceptable, or would it be better to give each library their own ID?

From a purely automated perspective, it doesn't really matter. But if I have to discuss this problem with a fellow human, I now have to clarify which library we are talking about as part of the conversation, so the identifier is really package:ID, not just the ID

kaniini:
I get what you're saying there.  In that case, I think a scanner should present it as two vulnerabilities, but prefixing it as subcomponent:ID.  But the scanner can do that itself, can't it?

oliverchang:
Thanks for the feedback! 

I completely agree that in most cases, there should only ever be one package specified in a single entry (and this is what we strongly recommend), and we really wanted to keep this. 

But for ease of adoption, I think we need to make a tradeoff to support interoperability with other schemas (such as CVE, GHSA) which do support (and rarely) have multiple packages listed (example: https://github.com/advisories/GHSA-wh77-3x4m-4q9g). 

We also got feedback from the Go vulnerability DB team, who wish to express that same package may have different import paths/names (vanity URLs, versioned imports) and they don't want to create separate entries for each, because it adds complexity for them. 

> I get what you're saying there. In that case, I think a scanner should present it as two vulnerabilities, but prefixing it as subcomponent:ID. But the scanner can do that itself, can't it?

I agree! The scanner would already know about what package it's scanning for, and then it just has to correlate that with what's listed in the vulnerability entry.

joshbressers:
I think this is all fair, I appreciate your explanation

schema.md
```diff
@@ -158,11 +161,64 @@ same database, the duplicate entry could be written using only the `id`,
...
+The `versions` field can enumerate a specific set of affected versions, and the
+`ranges` field can list ranges of affected versions, under a given defined
+ordering. **A version is considered affected if it lies within any one of the
+ranges or is listed in the versions list.**
+
+The `versions` list should - with one exception - always be present, to allow
+software to answer the question "is this specific version affected?" without
+having to contain code specific to every different ecosystem. The one exception
+is if the affected versions are valid SemVer 2.0 versions which can be
+accurately summarized by one or more non-overlapping SemVer ranges. In that
```
pombredanne, line 206: @oliverchang Singling out semver as being a perfect version scheme assumes that humans handling semver versions will never break semver. I think that this cannot hold at scale. 
 IMHO there is no reason to treat SEMVER specially and I explained some rationale in the related https://github.com/CVEProject/cve-schema/issues/87#issuecomment-904398866

Instead a version range may best treated as what it is: a hint for tools and humans to help form a proper `versions` enumeration.

oliverchang: Thanks for the feedback @pombredanne. I agree an explicit list of versions makes this unambiguous in all cases.

I think tooling/validators can help with making sure the SEMVER values are valid. One other reason why we need this is that in ecosystems like Go (which use/enforce Semver, it's infeasible to enumerate every possible version. Every commit can be a "pseudoversion" (valid semver) in Go.

pombredanne: Thank you for the clarification. This makes sense now. I had not internalized this aspect of Go modules versioning. That's actually pretty sleek and clean. It creates some extra work when trying to enumerate these when compared to other package types... but that's an issue for tools/db to deal with..!

PR #2, opened 2021-08-30
Clarification to required entry types for the "events" array.

oliverchang:
Change the requirement to an "introduced" event, as a single "fixed"
event is likely made in mistake. That is, the following means nothing is
affected:

```json
{
  "events": [
    {"fixed": "0.5"}
  ]
}
```


oliverchang:
Two different people have told me that the current requirement is confusing and can lead to bad entries getting generated.

inferno-chromium (approved):

PR #5, opened 2021-09-13
Add a JSON schema for validation.

oliverchang:
Includes validating:

- "repo" is required if range type is "GIT".
- "versions" is required if there are no SEMVER ranges. 
- There is at least once "introduced" event.


oliverchang:
@asraa can you please help with reviewing this as well? Thanks!

PR #8, opened 2021-09-23
Add a reference GHSA->OSV converter.

oliverchang:


tools/ghsa/convert_ghsa.py
```diff
@@ -0,0 +1,277 @@
...
+    cvss = ghsa.get('cvss', {})
+    cwes = ghsa.get('cwes', {}).get('nodes', [])
+
+    # Convert the grouped vulnerabilities in OSV range structures.
+    affected = []
+    for (ecosystem, name), vulns in package_to_vulns.items():
+        current = {
+            'package': {
+                'ecosystem': ecosystem,
+                'name': name,
```
meder, line 176: I assume GHSA uses same normalization rules as OSV?

meder (approved):

tools/ghsa/convert_ghsa.py
oliverchang: Thanks for catching this! Looks like for PyPI at least, GHSA does not perform normalization. I've fixed this.

PR #10, opened 2021-11-01
Add Debian ecosystem

oliverchang:


oliverchang:
@rsc what do you think of this general naming convention for Linux distro packages?

docs/schema.md
```diff
@@ -318,6 +318,7 @@ The defined ecosystems are:
 | `Maven` | The Maven Java package ecosystem. The `name` field is a Maven package name. |
 | `NuGet` | The NuGet package ecosystem. The `name` field is a NuGet package name. |
 | `Linux` | The Linux kernel. The only supported `name` is `Kernel`. |
+| `Debian` | The Debian package ecosystem; the `name` field is of the format `<lowercase distro codename>:<package name>`. For example, `buster:python3` refers to the "python3" package in Debian Buster. |
```
pombredanne, line 321: Is this intended to also support Ubuntu and Debian derivatives?

oliverchang: `Ubuntu` would be its own ecosystem. This is just a start and we'll need to add more later.

pombredanne: Thanks! And why would you need a `<lowercase distro codename>` for Debian? All packages live in a single unified pool that is not specific to a "distro codename" as in http://ftp.us.debian.org/debian/pool/main/f/ffmpeg/
I reckon debian publishes vulnerabilities with a distro info as in https://security-tracker.debian.org/tracker/source-package/ffmpeg but would make this required?

oliverchang: Thanks! I wasn't aware that packages are part of a unified pool. That makes this a lot simpler.

jeremiah: @oliverchang There has been some discussion about adoption of a database schema similar to the OSV ecosystem, but I don't know at this moment if there is consensus yet for adoption. We are aiming to determine that an to see if we have resources for an implementation.

oliverchang: @jeremiah that sounds awesome! Happy to answer any questions or help in any way we can to help with adoption.

May I ask what who you are referring to exactly when you say "we"?

rhertzog: May I suggest to update the description to say "the `name` is the name of the source package.". It makes it clearer to what we are referring to in Debian.

jeremiah: @oliverchang I use the term "we" very loosely here. :-) I am working with Freexian on Debian project funding and there is interest in funding work on various Debian infrastructure projects, including the Security Tracker. I'm hoping to coordinate discussion on possible adoption of the osv-schema and improvements to the process in general. 

Debian is a volunteer project as I'm sure you know so I don't speak on behalf of the project, just as someone who's very keen to support and improve Debian and Free Software in general.

oliverchang: @rhertzog Thanks for the feedback! Very happy to clarify this. However would "binary packages" be better here? We can still encode the source package elsewhere (e.g. the "ecosystem_specific" extension field). One of the intentions of this schema is to make it easier for vulnerability scanners to map against a list of packages/SBOM, and it seems like binary packages might be easier to consume? 

@jeremiah Thanks for clarifying! We would be very happy to get involved here and help any way we can to aid in adoption of this schema for Debian!

rhertzog: @oliverchang It's easy for a scanner to extract the source package name out of the dpkg metadata when scanning an installed system. And then binary package names are not stable over time, for example libraries typically embed the version encoded in SONAME in their binary package names (e.g. libidn12). Also you can have multiple binary packages for a single source package (and the sets of binary packages can also evolve over time) which would really complicate things here IMO. source package name is the most stable identifier in Debian (but then it's not necessarily unique either, for example we have multiple copies of postgresql and we embed the major version in the source package name).

pombredanne: @rhertzog I reckon that Debian tracks its security advisories at the source package level by release, but IMHO this creates an indirection that does not help actual users as they search, consume and install binary packages (that I can detect for instance with ScanCode.io and debian-inspector tools and ScanCode toolkit and lookup for vulnerabilities with VulnerableCode).

> (but then it's not necessarily unique either, for example we have multiple copies of postgresql and we embed the major version in the source package name).

Since the major version is embedded, the source package name is unique then? Just to be sure as it would be news to me if Debian package names were not guaranteed to be unique :angel:.And AFAIK the same is true for binary packages?

rhertzog: > this creates an indirection that does not help actual users as they search, consume and install binary packages

I agree with this but this indirection should be dynamically resolved by the tools for the benefits of the users. Changing the underlying data does not help except to introduce mistakes in the database because the binary package name for buster will not be the same as the one for bullseye.

> Since the major version is embedded, the source package name is unique then?

Yes... but I wanted to say is that the mapping from vulnerability in "foo" affects source package "foo-in-debian" is not necessarily a one-to-one mapping.

oliverchang: Sorry, I dropped the ball on this thread!

> but IMHO this creates an indirection that does not help actual users as they search, consume and install binary packages

+++1 to this. The intention is for this schema to be usable by generic tools without having any knowledge of ecosystem intricacies. This indirection could introduce additional complexity for users who may be consuming this data in contexts where it's not easy to get the source package mapping. 

> I agree with this but this indirection should be dynamically resolved by the tools for the benefits of the users. Changing the underlying data does not help except to introduce mistakes in the database because the binary package name for buster will not be the same as the one for bullseye.

This does sound like a great benefit for Debian. Each affected package also has an `ecosystem_specific` field that would be perfect for this. For specialized Debian tooling, they can read the source package name from this field also, and possibly also auto-populate / expand the actual binary packages that are affected.

rhertzog: What kind of generic tools are you referring too? How would they consume this data?

Do you expect Debian to do the mapping source to binary for all releases before injecting data in this new database? Do you know that this will greatly increase the size of the submitted data for dubious gain?

Shaheenkha1 (commented):
#87

PR #12, opened 2021-11-09
Rearrange repo attribute in example

ryru:
The repo attribute is not defined as property within the affected array but within an object of a ranges array.


oliverchang (approved):
thank you for fixing this!

PR #13, opened 2021-11-22
Update UVI to GSD

athix:
We recently renamed the Universal Vulnerability Identifier (UVI) project to Global Security Database (GSD)


oliverchang (approved):
thanks!

PR #15, opened 2021-12-05
Rename UVI to GSD within Github Pages

athix:
Follow up to #13, noticed that I missed the Github pages reference.


oliverchang (approved):
Thanks!

PR #16, opened 2021-12-14
Add top-level database_specific field

chrisbloom7:
This adds a top-level `database_specific` field to allow source databases to embed custom metadata that is relevant to the entire vulnerability rather than specific affected products.


oliverchang (approved):
Thanks! This looks pretty straightforward to me. I'll let others chime in here before merging.

rsc:
I agree. This seems fine.

rsc (approved):

PR #17, opened 2021-12-14
Add a schema_version field for tracking which version of OSV Schema a vulnerability was produced against

chrisbloom7:
It will be helpful as this schema evolves to be able to distinguish between which version a particular vulnerability was exported against. This provides downstream consumers with the ability to decide how to import a vulnerability as well as provide some protection against any future breaking schema changes.

There is currently no standard or recommended best practice for versioning JSON schemas, so I made two decisions in this PR that perhaps require further discussion:

1. Naming the field `schema_version` versus `version` - Since we're dealing with affected product versions it seemed better to be explicit about the purpose of this field so as not to introduce any ambiguity or chance for misinterpretation.
2. Using SemVer format for indicating the schema version - It seems like the OSV Schema is following this format for versioning, but please correct me if I'm wrong. An alternative format named [SchemaVer](https://snowplowanalytics.com/blog/2014/05/13/introducing-schemaver-for-semantic-versioning-of-schemas/#schemaver) has been proposed, but it's not widely adopted and since this documentation already discusses SemVer at length for affected product versions it seemed like we should avoid introducing a competing format.

### DIFF
--- a/docs/schema.md
+++ b/docs/schema.md
@@ -56,6 +56,7 @@ A JSON Schema for validation is also available
 
 ```json
 {
+	"schema_version": string,
 	"id": string,
 	"modified": string,
 	"published": string,
@@ -102,6 +103,22 @@ absolutely must be shared between databases, leaving customizations to the
 "ecosystem_specific" and "database_specific" blocks (see below)
 
 # Field Details
+
+## schema_version field
+
+```json
+{
+	"schema_version": string
+}
+```
+
+The `schema_version` field is used to indicate which version of the OSV schema
+a particular vulnerability was exported with. This can help consumer
+applications decide how to import the data for their own systems and offer some
+protection against future breaking changes. The value should be a string using
+the [SchemaVer](https://snowplowanalytics.com/blog/2014/05/13/introducing-schemaver-for-semantic-versioning-of-schemas/)
+format of `<MODEL>-<REVISION>-<ADDITION>`. If no value is specified, it should
+be assumed to be `1-0-0`, matching version 1.0.0 of the OSV Schema.
 
 ## id, modified fields
 
--- a/docs/schema.md
+++ b/docs/schema.md
@@ -113,12 +113,15 @@ absolutely must be shared between databases, leaving customizations to the
 ```
 
 The `schema_version` field is used to indicate which version of the OSV schema
-a particular vulnerability was exported with. This can help consumer
-applications decide how to import the data for their own systems and offer some
-protection against future breaking changes. The value should be a string
-matching the OSV Schema version, which follows the [SemVer 2.0.0](https://semver.org)
-format, with no leading "v" prefix. If no value is specified, it should
-be assumed to be `1.0.0`, matching version 1.0 of the OSV Schema.
+a particular vulnerability was exported with. This can help consumer applications
+decide how to import the data for their own systems and offer some protection
+against future breaking changes. The value should be a string matching the OSV
+Schema version, which follows the [SemVer 2.0.0](https://semver.org) format, with
+no leading "v" prefix. If no value is specified, it should be assumed to be `1.0.0`,
+matching version 1.0 of the OSV Schema. Clients can assume that new minor and patch
+versions of the schema only add new fields, without changing the meaning of old
+fields, so that a client that knows how to read version 1.2.0 can process data
+identifying as schema version 1.3.0 by ignoring any unexpected fields.
 
 ## id, modified fields
 
--- a/docs/schema.md
+++ b/docs/schema.md
@@ -115,10 +115,10 @@ absolutely must be shared between databases, leaving customizations to the
 The `schema_version` field is used to indicate which version of the OSV schema
 a particular vulnerability was exported with. This can help consumer
 applications decide how to import the data for their own systems and offer some
-protection against future breaking changes. The value should be a string using
-the [SchemaVer](https://snowplowanalytics.com/blog/2014/05/13/introducing-schemaver-for-semantic-versioning-of-schemas/)
-format of `<MODEL>-<REVISION>-<ADDITION>`. If no value is specified, it should
-be assumed to be `1-0-0`, matching version 1.0.0 of the OSV Schema.
+protection against future breaking changes. The value should be a string
+matching the OSV Schema version, which follows the [SemVer 2.0.0](https://semver.org)
+format, with no leading "v" prefix. If no value is specified, it should
+be assumed to be `1.0.0`, matching version 1.0 of the OSV Schema.
 
 ## id, modified fields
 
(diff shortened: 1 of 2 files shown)


docs/schema.md
```diff
@@ -103,6 +104,22 @@ absolutely must be shared between databases, leaving customizations to the
...
+{
+	"schema_version": string
+}
+```
+
+The `schema_version` field is used to indicate which version of the OSV schema
+a particular vulnerability was exported with. This can help consumer
+applications decide how to import the data for their own systems and offer some
+protection against future breaking changes. The value should be a string using
+the [SchemaVer](https://snowplowanalytics.com/blog/2014/05/13/introducing-schemaver-for-semantic-versioning-of-schemas/)
```
rsc: I'm somewhat concerned about adopting something that only exists in a blog post marked as a "draft". Do any other schemas use this format? The fact that below it says that 1-0-0 really means 1.0.0 suggests that maybe we should just use SemVer. We can easily say here what the meanings of 'major', 'minor', and 'patch' are, and then we avoid the dependency on a 7-year-old blog post. (I don't really agree with the blog post's claim that SemVer is fundamentally different than this thing they call SchemaVer. Why not just use SemVer and be done? Certainly in 2021 the case for SemVer is even stronger than it was in 2014.)

oliverchang: I think you might be looking at an outdated commit @rsc :) The latest changes just use SemVer. 

Are you opposed at all in principle to adding this field? 

I can see the argument that given all our format changes are intended to be backwards compatible, we can defer adding a version field until we **really** need to break something with a v2.

rsc: If we are going to add a version field, we need to define how old clients should react to seeing newer versions. The old SchemaVer was bad for using an obscure format but good for defining that well. I agree with using SemVer in the new commit but then we should add more text.

That is, we should add something like:

> If no value is specified, it should be assumed to be `1.0.0`, matching version 1.0 of the OSV Schema.
> Clients can assume that new minor and patch versions of the schema
> only add new fields, without changing the meaning of old fields,
> so that a client that knows how to read version 1.2.0 can process data
> identifying as schema version 1.3.0 by ignoring any unexpected fields.

Assuming that is what we mean. I don't know exactly what we mean. 

@chrisbloom7, how did you intend clients to interpret this field?

chrisbloom7: @rsc thanks for weighing in on this. We are planning on exporting a large group of vulnerabilities in the near future, and then continuing to add to that collection as new vulnerabilities come in. The initial collection will be exported with version 1.0 of the spec, and we plan on keeping up with changes to the spec over time such that those exported say 1 year from now will follow whatever version is published at that time. But, someone may decide to make a change to one of the older exported files and then we will need to import it, and at that point it will be useful to know what version of the spec a file was generated with so we can import it properly regardless of the current spec. We haven't built that tooling yet, but I imagine the process would involve checking the version the file was exported with and mapping that to the importer that supports that version. It's likely that only the major version would matter to us in this case, but more nuanced detection on the minor and patch levels can't be ruled out entirely.

I'll also add this note from our meeting with Oliver:

> Version of the schema being “emitted”, i.e. this advisory is { schema_version: “1.0” }, so that as the schema evolves we can guarantee that parsers won’t break. This stems from our own experience emitting advisory API schemas; a year or two or three down the road there’ll be a breaking change to the OSV schema.

chrisbloom7: (Oops, I just noticed you also commented on that section from the doc and so were familiar with it already)

rsc: Thanks for the context. Oliver summarized your meeting, and to be clear I'm completely supportive of adapting OSV to help GitHub's use cases. I'm just trying to understand what clients can do with this field, if anything at all. I think if we add the field we should make clear how to use it.

Obviously if a newer client reads an old file, it knows about the old version and can do whatever is right without help from us. It sounds like that's the case you are intending, and that's taken care of.

In the case where an older client reads a new file, I think it's important to say how the client should handle the change. Specifically:
- Should the old client refuse to read the record? 
- Should it assume the parts of the record it recognizes have the same meaning as in the older version it was programmed against? 
- Something else?

I didn't like SchemaVer's syntax, but I did like that it tried to answer these questions in a way that clients could use.

I think we could reasonably adopt the rule that we put a full 3-number version (1.0.0) on the top of the spec, that we bump the patch number when we edit the text without changing the meaning of the fields, that we bump the minor number when we add new fields but leave the existing ones' meanings unchanged, and that we bump the major number when we make an actual breaking change so that fields mean something different now. If we do that, then clients would refuse to read new major versions but plow ahead on new minor versions and new patch versions. Does that seem reasonable to you? (If this isn't a case GitHub cares about, then Oliver and I can just figure it out among ourselves I guess.)

If we do that, then I think the text I posted earlier works:

> If no value is specified, it should be assumed to be 1.0.0, matching version 1.0 of the OSV Schema.
> Clients can assume that new minor and patch versions of the schema
> only add new fields, without changing the meaning of old fields,
> so that a client that knows how to read version 1.2.0 can process data
> identifying as schema version 1.3.0 by ignoring any unexpected fields.

chrisbloom7: > I'm just trying to understand what clients can do with this field, if anything at all. I think if we add the field we should make clear how to use it. 

> Obviously if a newer client reads an old file, it knows about the old version and can do whatever is right without help from us. It sounds like that's the case you are intending, and that's taken care of.

> In the case where an older client reads a new file, I think it's important to say how the client should handle the change.

I see, thanks for clarifying. I agree that will be helpful info to have in the spec to help consumers prepare for future changes.

> I think we could reasonably adopt the rule that we put a full 3-number version (1.0.0) on the top of the spec, that we bump the patch number when we edit the text without changing the meaning of the fields, that we bump the minor number when we add new fields but leave the existing ones' meanings unchanged, and that we bump the major number when we make an actual breaking change so that fields mean something different now. If we do that, then clients would refuse to read new major versions but plow ahead on new minor versions and new patch versions. Does that seem reasonable to you? (If this isn't a case GitHub cares about, then Oliver and I can just figure it out among ourselves I guess.)

That sounds perfect, and accomplishes what we're after.

> If no value is specified, it should be assumed to be 1.0.0, matching version 1.0 of the OSV Schema.
> Clients can assume that new minor and patch versions of the schema
> only add new fields, without changing the meaning of old fields,
> so that a client that knows how to read version 1.2.0 can process data
> identifying as schema version 1.3.0 by ignoring any unexpected fields.

Yes, this makes sense. I'll update the docs

oliverchang (approved):
This LGTM! Will leave this to Russ to take another look.

oliverchang:
Actually, could you please edit the version at the top of the spec to be "1.1.0" to be consistent? We can merge this PR last.

rsc (approved):
Looks great, thanks for updating the PR.

PR #18, opened 2021-12-14
Drop versions requirement for `ECOSYSTEM` ranges

chrisbloom7:
This resolves a semantic issue: Not all ecosystems that recommend using SemVer 2.0 for versioning actually enforce that format on version numbers, but it may still be possible for the ecosystem to determine which versions are affected based on a range of SemVer-like versions. To address this, after speaking with @oliverchang, we are proposing dropping the validation of a `versions` requirement for `ECOSYSTEM` ranges, and changing the documentation to state that versions are strongly recommended for when using `ECOSYSTEM` ranges but not required. They are still required for `GIT` ranges.

### DIFF
--- a/docs/schema.md
+++ b/docs/schema.md
@@ -271,7 +271,7 @@ ranges, they may be encoded using the `ranges` field with an appropriate type
 `versions` list. Tooling and infrastructure such as https://osv.dev are able to
 expand these ranges for supported ecosystems into the `versions` list for easier
 consumption. Products or ecosystems that do not use version identifiers that can
-be represented as ranges **must** include the enumerated `versions` list instead.
+be represented as ranges must include the enumerated `versions` list instead.
 
 ### affected[].package field
 
--- a/docs/schema.md
+++ b/docs/schema.md
@@ -265,17 +265,13 @@ ranges or is listed in the versions list.**
 The `versions` list should - with one exception - always be present, to allow
 software to answer the question "is this specific version affected?" without
 having to contain code specific to every different ecosystem. The one exception
-is if the affected versions are valid SemVer 2.0 versions which can be
-accurately summarized by one or more non-overlapping SemVer ranges. In that
-case, the SemVer ranges can be listed instead, in entries in the `ranges` field
-with type `SEMVER` (see below). In this case, the SemVer ranges act as a kind of
-compact form of a larger `versions` list. Ecosystems that do not use SemVer
-identifiers or that order versions differently from SemVer must include the
-enumerated `versions` list, although they can also add ranges of type
-`ECOSYSTEM` for additional context.
-
-In short, each object in the `affected` array must contain either a non-empty
-`versions` list or at least one range in the `ranges` list of type `SEMVER`.
+is if the affected versions can be accurately summarized by one or more non-
+overlapping ranges. In that case, the ranges can be listed instead, in entries
+in the `ranges` field with an appropriate type (see below). In this case, the
+ranges act as a kind of compact form of a larger `versions` list. Products or
+ecosystems that do not use version identifiers that can be represented as ranges
+should include the enumerated `versions` list instead, though they can also add
+ranges of the appropriate type for additional context.
 
 ### affected[].package field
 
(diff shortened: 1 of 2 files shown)


docs/schema.md
```diff
@@ -265,17 +265,13 @@ ranges or is listed in the versions list.**
 The `versions` list should - with one exception - always be present, to allow
```
oliverchang, line 265: This paragraph seems a little inconsistent now, so let's reword this like:

```
The `versions` list is generally recommended to always be present, to allow
software to easily answer the question "is this specific version affected?" without
having to contain code specific to every different ecosystem. If the affected
versions can be accurately summarized by one or more non-overlapping
ranges, they may be encoded using the `ranges` field with an appropriate
type (see below).
In this case, the ranges act as a kind of compact form of a larger `versions` list. Tooling
and infrastructure such as https://osv.dev are able to expand these ranges for
supported ecosystems into the `versions` list for easier consumption. Products or
ecosystems that do not use version identifiers that can be represented as ranges
**must** include the enumerated `versions` list instead.
```

docs/schema.md
```diff
@@ -373,9 +372,11 @@ an explicitly enumerated `versions` list, because `GIT` range inclusion queries
 cannot be answered without access to a copy of the underlying Git repository.
 
 Again, it is important to note that to allow portable (non-ecosystem-specific)
```
oliverchang, line 374: I think let's just remove this entire paragraph, because the point it's making here is no longer as strong.

oliverchang (commented):
thanks for this @chrisbloom7 !

Just some comments on wording, but I'm happy with the general loosening of this requirement.

rsc:
The goal of the type here is to give readers enough information to be able to algorithmically answer the question "is the version I see in my code in the affected range?". Different ecosystems have different answers for this. We started simple with just SEMVER, GIT, and the wildcard ECOSYSTEM, but it would be OK to define a new type for each ecosystem, provided there is a well-defined algorithm for answering the "version in range?" question.

Loosening SEMVER doesn't help - what if two different ecosystems loosen semver in different ways?

Dropping the ECOSYSTEM enumeration requirement may be OK as long as package.ecosystem is a known ecosystem that all processors are expected to implement a "version in range?" for. But instead of having to define the ones where ECOSYSTEM can omit the version list, it makes more sense to me to just add more 'type' fields, each linked to a well-defined spec for version syntax and the meaning of < and <=.

oliverchang:
> The goal of the type here is to give readers enough information to be able to algorithmically answer the question "is the version I see in my code in the affected range?". Different ecosystems have different answers for this. We started simple with just SEMVER, GIT, and the wildcard ECOSYSTEM, but it would be OK to define a new type for each ecosystem, provided there is a well-defined algorithm for answering the "version in range?" question.
> 
> Loosening SEMVER doesn't help - what if two different ecosystems loosen semver in different ways?
> 
> Dropping the ECOSYSTEM enumeration requirement may be OK as long as package.ecosystem is a known ecosystem that all processors are expected to implement a "version in range?" for. But instead of having to define the ones where ECOSYSTEM can omit the version list, it makes more sense to me to just add more 'type' fields, each linked to a well-defined spec for version syntax and the meaning of < and <=.

I believe there will be a strong mapping from `package.ecosystem` to the version scheme denoted by `ECOSYSTEM`. Having to define each one separately (where the name is either the same or slightly different -- e.g. "PyPI" vs "PEP-440") seems unnecessarily cumbersome to me. 

Tooling like https://osv.dev and https://github.com/nexB/univers will also make this easier to consume in general.

rsc:
OK, dropping the enumeration requirement for 'ECOSYSTEM' sounds reasonable then. Thanks.
I agree with Oliver's clarifications to the text.

chrisbloom7:
@oliverchang thanks for the change suggestions on wording in the docs. I've incorporated them as requested.

docs/schema.md
```diff
@@ -262,20 +262,16 @@ The `versions` field can enumerate a specific set of affected versions, and the
...
+The `versions` list is generally recommended to always be present, to allow
+software to easily answer the question "is this specific version affected?"
+without having to contain code specific to every different ecosystem. If the
+affected versions can be accurately summarized by one or more non-overlapping
+ranges, they may be encoded using the `ranges` field with an appropriate type
+(see below). In this case, the ranges act as a kind of compact form of a larger
+`versions` list. Tooling and infrastructure such as https://osv.dev are able to
+expand these ranges for supported ecosystems into the `versions` list for easier
+consumption. Products or ecosystems that do not use version identifiers that can
+be represented as ranges **must** include the enumerated `versions` list instead.
```
rsc, line 274: None of the other 'must' in the doc are bold; please unbold this one.

rsc (approved):
Will leave for Oliver. Thanks!

docs/schema.md
chrisbloom7: That formatting was based on @oliverchang's [suggestion](https://github.com/ossf/osv-schema/pull/18#discussion_r769186679) above. Happy to go either way. Oliver?

oliverchang: Let's remove this to be consistent as @rsc pointed out. Thanks!

chrisbloom7: Done!

oliverchang (approved):
thanks!

PR #19, opened 2021-12-16
add Hex to the ecosystems for our erlang advisories

katblag:
This adds [Hex](https://hex.pm/) to the documented list of defined ecosystems. The description is the wording taken directly from their website, but open to feedback if you'd like it to follow a specific pattern.

There are no validations on the ecosystems defined in the schema docs, so the validation JSON has not been updated.


oliverchang (approved):
Thanks!

rsc (approved):

PR #20, opened 2022-01-01
Invalid OSV json example

ryru:
In my understanding the OSV json is an invalid specification version 1.0.0 example, since it lakes a `versions` field and does not have a range of a SemVer 2.0 type:

> In short, each object in the `affected` array must contain either a non-empty
> `versions` list or at least one range in the `ranges` list of type `SEMVER`.

Dropping the `versions` requirement in #18 makes it valid although only for parsers that support the specification as of 1.1.0. Or have I missed something?


oliverchang (approved):
Thank you! Indeed we forgot to update the examples here.

PR #21, opened 2022-01-07
Add optional top-level severity field

chrisbloom7:
This adds a top-level `severity` array field field to capture vulnerability severity rankings such as [CVSS](https://www.first.org/cvss/). Each severity is an object with a `type` enum and a `score`.

Initially supported types are `CVSS_V3` and `PLAIN_TEXT`, described in the schema documentation.

### DIFF
--- a/docs/schema.md
+++ b/docs/schema.md
@@ -252,8 +252,8 @@ display sites to unnecessary vulnerabilities.)
 	"severity": [ {
 		"type": string,
 		"score": string
-	}
-} ]
+	} ]
+}
 ```
 
 The `severity` field is a JSON array that allows generating systems to describe
--- a/docs/schema.md
+++ b/docs/schema.md
@@ -65,6 +65,10 @@ A JSON Schema for validation is also available
 	"related": [ string ],
 	"summary": string,
 	"details": string,
+	"cvss": {
+		"score": string,
+		"version": string
+	},
 	"affected": [ {
 		"package": {
 			"ecosystem": string,
@@ -240,6 +244,25 @@ as stripping raw HTML and links that do not start with http:// or https://.
 Databases are encouraged not to include those in the first place. (The goal is
 to balance flexibility of presentation with not exposing vulnerability database
 display sites to unnecessary vulnerabilities.)
+
+## cvss field
+
+```json
+{
+	"cvss": {
+		"score": string,
+		"version": string
+	}
+}
+```
+
+The `cvss` field represents the unique characteristics and severity of the
+vulnerability using the [Common Vulnerability Scoring System notation](https://www.first.org/cvss/).
+The field is specified as an object with a string property `score`, and an
+optional string property `version` denoting the version of CVSS that was used to
+calculate the score, which can be helpful for determining how the score was
+derived. The `cvss` field itself is optional, but if present it must contain
+at least a `score` property.
 
 ## affected fields
 
--- a/docs/schema.md
+++ b/docs/schema.md
@@ -65,10 +65,10 @@ A JSON Schema for validation is also available
 	"related": [ string ],
 	"summary": string,
 	"details": string,
-	"cvss": {
-		"score": string,
-		"version": string
-	},
+	"severity": [ {
+		"type": string,
+		"score": string
+	} ],
 	"affected": [ {
 		"package": {
 			"ecosystem": string,
@@ -245,24 +245,38 @@ Databases are encouraged not to include those in the first place. (The goal is
 to balance flexibility of presentation with not exposing vulnerability database
 display sites to unnecessary vulnerabilities.)
 
-## cvss field
+## severity field
 
 ```json
 {
-	"cvss": {
-		"score": string,
-		"version": string
+	"severity": [ {
+		"type": string,
+		"score": string
 	}
-}
+} ]
 ```
 
-The `cvss` field represents the unique characteristics and severity of the
-vulnerability using the [Common Vulnerability Scoring System notation](https://www.first.org/cvss/).
-The field is specified as an object with a string property `score`, and an
-optional string property `version` denoting the version of CVSS that was used to
-calculate the score, which can be helpful for determining how the score was
-derived. The `cvss` field itself is optional, but if present it must contain
-at least a `score` property.
+The `severity` field is a JSON array that allows generating systems to
+describe the severity of a vulnerability using one or more scoring
+methods. Each `severity` item is a JSON object specifying a `type` and
+`score` property, described below.
+
+### severity[].type field
+
+The `severity[].type` property must be one of the types defined below,
+which describe how the `associated` score was derived. This allows
+processing systems to decide how to handle each severity.
+
+| Severity Type | Score Description |
+| --------- | ----------- |
+| `CVSS_V3` | A computed score representing the unique characteristics and severity of the vulnerability using a version of the [Common Vulnerability Scoring System notation](https://www.first.org/cvss/) that is >= 3.0 and < 4.0. |
+| `PLAIN_TEXT` | A human-derived and human-readable rating of the severity, based on interpretation of available vulnerability information. |
+| Your severity type here. | [Send us a PR](https://github.com/ossf/osv-schema/compare). |
+
+### severity[].score field
+
+The `severity[].score` property is a string ranking the severity based on the
+selected `severity[].type`, as described above.
 
 ## affected fields
 
--- a/docs/schema.md
+++ b/docs/schema.md
@@ -256,27 +256,25 @@ display sites to unnecessary vulnerabilities.)
 } ]
 ```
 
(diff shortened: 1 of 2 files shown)


docs/schema.md
```diff
@@ -65,6 +65,10 @@ A JSON Schema for validation is also available
 	"related": [ string ],
 	"summary": string,
 	"details": string,
+	"cvss": {
```
oliverchang, line 68: I wonder if we should just be more generic here (beyond just CVSS) to allow for future extensibility. 

i.e. something like:

```json
"severity": {
  "type": "CVSSv3",
  "score": "...."
}
```

oliverchang (commented):
Thanks! Just a comment on making this more generic for future extensibiilty. @rsc WDYT?

docs/schema.md
chrisbloom7: Oh, that's interesting! That's certainly more extensible. What do you think about it being an array in that case. Then you could do something like:

```
"severity": [
  {
    "type": "CVSSv3",
    "score": "...."
  },
  {
    "type": "Curation",
    "score": "Severe"
  },
  {
    "type": "some-ecosystem-specific-thing",
    "score": "xyz"
  }
]
```

This would solve another problem we were facing which was where to put a human readable `severity` field - we'd add it here alongside cvss.

oliverchang: Indeed that's a great idea! An array would be much more flexible and expressive. 

My only concern though is with the "ecosystem specific" example you have there. Ideally all the ecosystem specific stuff should be confined to the "ecosystem_specific" field instead.

chrisbloom7: Yeah, that's fair. It was only a contrived example for demonstration. If we go this route, it seems reasonable to have some predefined "types" as we do with `references[].type`, the proposed `credits[].contact[].type`, etc., to make these more consistent and more easily processesable by consuming systems. A `CVSS_V3` type of course, and maybe "Curation" becomes `CURATED` or `PLAIN_TEXT` or something similar to denote it's human derived and human readable. Any others that you think we should include up front?

chrisbloom7: I've converted the `cvss` field to a `severity` array field. I'll updated the PR description to match.

chrisbloom7:
I've converted the `cvss` field to a `severity` array field as discussed and updated the PR description to match.

docs/schema.md
```diff
@@ -241,6 +245,39 @@ Databases are encouraged not to include those in the first place. (The goal is
...
 
+## severity field
+
+```json
+{
+	"severity": [ {
+		"type": string,
+		"score": string
+	}
+} ]
```
oliverchang, line 256: The } and ] should be swapped here. To format this consistently, the ] should go after the } on line 255 instead.

validation/schema.json
```diff
@@ -40,6 +40,29 @@
...
+      "type": "array",
+      "items": {
+        "type": "object",
+        "properties": {
+          "type": {
+            "type": "string",
+            "enum": [
+              "CVSS_V3",
+              "PLAIN_TEXT",
+              "OTHER"
```
oliverchang, line 53: OTHER is not in the documentation. remove?

docs/schema.md
```diff
@@ -241,6 +245,39 @@ Databases are encouraged not to include those in the first place. (The goal is
...
+### severity[].type field
+
+The `severity[].type` property must be one of the types defined below,
+which describe how the `associated` score was derived. This allows
+processing systems to decide how to handle each severity.
+
+| Severity Type | Score Description |
+| --------- | ----------- |
+| `CVSS_V3` | A computed score representing the unique characteristics and severity of the vulnerability using a version of the [Common Vulnerability Scoring System notation](https://www.first.org/cvss/) that is >= 3.0 and < 4.0. |
+| `PLAIN_TEXT` | A human-derived and human-readable rating of the severity, based on interpretation of available vulnerability information. |
```
oliverchang, line 273: Could we define this a bit more strictly? Is the intention to have a human readable vulnerability metric, i.e. "Critical", "High", "Medium", "Low"? If so, I think it would be better to state these here so that these entries are consistent across all vuln DBs.

oliverchang (commented):
Nice!! I really like this.

rsc:
Severity is tricky because it's very difficult to declare severity outside of context. This is particularly true of open-source libraries, where the severity depends almost entirely on how exactly the code is used. Severity is much easier to understand in whole systems like, say, a MacOS distribution. Even so, I understand the pressure to add it, and I won't object to the concept.

That said, a goal for this format is to be machine-readable and machine-understandable - in contrast to earlier versions of CVE that were basically English text. I am worried about "PLAIN\_TEXT" being a bit of a regression on that front. It reminds me of the human-readable-only 'version' field in earlier CVE schemas. I'd be a lot more comfortable if all the text that can only be parsed by humans were limited to the description. If we start with just CVSS\_V3 as a severity type, can we leave PLAIN\_TEXT out? 

Can we commit to the severity field only containing machine-parseable severity taxonomies?

docs/schema.md
chrisbloom7: Oops, thank you! Fixing

validation/schema.json
chrisbloom7: Oh, huh. Not sure what I had in mind when adding that. Will remove it.

chrisbloom7:
> a goal for this format is to be machine-readable and machine-understandable

Thanks for calling that out, it's useful to keep in mind. Maybe `PLAIN_TEXT` is not the best name for that type, or maybe I could have worded the description of it better. The distinction I was trying to make was that whereas CVSS scores are derived from a formula, a vulnerability can also be assigned a grade by the reporting entity that's based on multiple factors but not in a way that is necessarily formulaic. Ultimately a human is usually making that call based on intuition and/or reason. We could do as @oliverchang suggested and ensure the score for that type was limited to one of "CRITICAL", "HIGH", "MEDIUM", "LOW" which would ensure they were consistent and therefore more machine friendly while still having more value to a casual human consumer than a computed score might. Would that change your opinion of including it? I'm open to suggestions on a better name and description in that case. If not, we're OK moving forward with support for just the `CVSS_V3` type for now.

oliverchang:
> > a goal for this format is to be machine-readable and machine-understandable
> 
> Thanks for calling that out, it's useful to keep in mind. Maybe `PLAIN_TEXT` is not the best name for that type, or maybe I could have worded the description of it better. The distinction I was trying to make was that whereas CVSS scores are derived from a formula, a vulnerability can also be assigned a grade by the reporting entity that's based on multiple factors but not in a way that is necessarily formulaic. Ultimately a human is usually making that call based on intuition and/or reason. We could do as @oliverchang suggested and ensure the score for that type was limited to one of "CRITICAL", "HIGH", "MEDIUM", "LOW" which would ensure they were consistent and therefore more machine friendly while still having more value to a casual human consumer than a computed score might. Would that change your opinion of including it? I'm open to suggestions on a better name and description in that case. If not, we're OK moving forward with support for just the `CVSS_V3` type for now.

+1 to `PLAIN_TEXT` being not a good name if we go ahead with this.

A `"QUALITATIVE"` severity type with the well defined "CRITICAL", "HIGH", "MEDIUM", "LOW" values seems OK to me. These values can still be machine understandable if the values this field can have is consistent and well defined. @rsc do you think that would be OK?

rsc:
For the CRITICAL/HIGH/MEDIUM/LOW it would still be good to have a sense of what these mean, to keep the entries precise.
More generally each type we define should be backed by a definition, either in this spec or linked to from this spec.
Is there some agreed-upon industry-wide meaning for these? (Somehow I expect not.)
Are there databases that define their own meanings for these?
If so, maybe we should have a type named for the database that defines the scale, 
or perhaps add the field to database\_specific?

chrisbloom7:
@rsc I don't disagree that given the qualitative nature of this field it could easily be categorized as database specific since it depends on who is making that determination and what they feel is important to consider. I initially made the argument to my team that we should just place it in the new top-level `database_specific` field, but then when we pivoted on this PR to use an array of different scores it seemed like it could fit there as well. Let's take this qualitative type off the table for a moment since we agree it is difficult to define in a cross-platform way.

If we look at this as just an array of _quantitative_ severity scores, opening with just a CVSS v3 type, any concerns on moving forward that way? (And with the assumption that I will clean up the docs to match before final sign off)

oliverchang (approved):

rsc:
The new one with just CVSS_V3 looks like a good start. 
Thanks for putting "quantitative" in the bottom table entry.

Not for this PR, but one thing I would support in the future
if it were valuable to GH or others is some standard enumeration of
vulnerability categories, like "denial of service", "crash",
"remote code execution", and so on. 
Listing the worst-case possibilities in that form might help users
adapt their response appropriately.

rsc (approved):

PR #22, opened 2022-01-07
Add credits field

chrisbloom7:
This proposes a top-level credits field. We learned through internal research projects that credits are an important form of currency in the vulnerability community, and often a strong motivator for identifying or fixing vulnerabilities. It provides the benefit of name recognition as well as acts as a resume of sorts for vulnerability researchers. In order to proved proper credit, it's important that contributors are able to be represented with both a name as well as unique identifiers or ways to link back to other work they have done.

Because of this, we chose to implement credits as a list of objects rather than strings for a few reasons:

1. Contributors may not want their personal email addresses published, so we don't want to rely on simple lists of "First Last <email...>" values.
2. Different source databases may use different identifiers to represent contributing users
3. Contributors may wish to include multiple contact methods, or none
4. Processing systems may wish to process contact information dynamically

With those limitations in mind, we can define `credits` similar to how `references` are defined.

### DIFF
--- a/docs/schema.md
+++ b/docs/schema.md
@@ -628,9 +628,9 @@ The known reference `type` values are:
 }
 ```
 
-The `credits` field is a JSON array providing a way to give credit where credit
-is due related to the discovery, confirmation, patch, or other events in the life
-cycle of a vulnerability.
+The `credits` field is a JSON array providing a way to give credit for the
+discovery, confirmation, patch, or other events in the life cycle of a
+vulnerability.
 
 Each of the objects in the `credits` array must contain at minimum a `name` field
 specifying the name of the individual or entity being credited, using whatever
@@ -666,7 +666,6 @@ as necessary and appropriate.
 | `GITHUB` | A valid, plain-text GitHub username for the credited. |
 | `TWITTER` | A valid, plain-text Twitter username, with or without the leading `@` symbol, for the credited. |
 | `URL` | A fully-qualified plain-text URL |
-| `OTHER` | A plain-text value that is not represented by any other available type and is self-evident in its usage as a means of contacting the credited individual or entity. |
 | Your contact type here. | [Send us a PR](https://github.com/ossf/osv-schema/compare). |
 
 #### credits[].contact[].value field
--- a/docs/schema.md
+++ b/docs/schema.md
@@ -90,10 +90,7 @@ A JSON Schema for validation is also available
 	} ],
 	"credits": [ {
 		"name": string,
-		"contact": [ {
-			"type": string,
-			"value": string
-		} ]
+		"contact": [ string ]
 	} ],
 	"database_specific": { see description }
 }
@@ -620,10 +617,7 @@ The known reference `type` values are:
 {
 	"credits": [ {
 		"name": string,
-		"contact": [ {
-			"type": string,
-			"value": string
-		} ],
+		"contact": [ string ],
 	} ]
 }
 ```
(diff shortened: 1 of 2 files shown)


validation/schema.json
```diff
@@ -218,6 +218,45 @@
...
+              "type": "object",
+              "properties": {
+                "type": {
+                  "type": "string",
+                  "enum": [
+                    "EMAIL",
+                    "GITHUB",
+                    "OTHER",
+                    "TWITTER",
+                    "URL"
```
chrisbloom7, line 241: These are just the types that came off the top of my head. Happy to expand this list with any other initial values that you think are worth including.

docs/schema.md
```diff
@@ -607,6 +614,66 @@ The known reference `type` values are:
...
+as necessary and appropriate.
+
+| Contact Type | Value Description |
+| --------- | ----------- |
+| `EMAIL` | A valid, fully qualified, plain-text email adddress at which the credited can be reached. |
+| `GITHUB` | A valid, plain-text GitHub username for the credited. |
+| `TWITTER` | A valid, plain-text Twitter username, with or without the leading `@` symbol, for the credited. |
+| `URL` | A fully-qualified plain-text URL |
+| `OTHER` | A plain-text value that is not represented by any other available type and is self-evident in its usage as a means of contacting the credited individual or entity. |
+| Your contact type here. | [Send us a PR](https://github.com/ossf/osv-schema/compare). |
```
chrisbloom7, line 670: Mimicking the last entry of the ecosystems table

docs/schema.md
```diff
@@ -607,6 +614,66 @@ The known reference `type` values are:
...
+		"name": string,
+		"contact": [ {
+			"type": string,
+			"value": string
+		} ],
+	} ]
+}
+```
+
+The `credits` field is a JSON array providing a way to give credit where credit
```
oliverchang, line 631: nit: just "give credit for the discovery, confirmation, patch..." is more concise?

docs/schema.md
```diff
@@ -607,6 +614,66 @@ The known reference `type` values are:
...
+allows processing systems to interpret and display contact information dynamically
+as necessary and appropriate.
+
+| Contact Type | Value Description |
+| --------- | ----------- |
+| `EMAIL` | A valid, fully qualified, plain-text email adddress at which the credited can be reached. |
+| `GITHUB` | A valid, plain-text GitHub username for the credited. |
+| `TWITTER` | A valid, plain-text Twitter username, with or without the leading `@` symbol, for the credited. |
+| `URL` | A fully-qualified plain-text URL |
+| `OTHER` | A plain-text value that is not represented by any other available type and is self-evident in its usage as a means of contacting the credited individual or entity. |
```
oliverchang, line 669: I would prefer we omit "OTHER", given that we have a "URL" field. Having a free form "OTHER" field where anything goes isn't very actionable for consumers. Ideally we'd want every field in this schema to be well formed and easily parseable (apart from the English vulnerability description text).

oliverchang (commented):
Thanks! This generally looks good to me, with 2 minor comments.

@rsc WDYT?

rolandshoemaker:
From the Go side: this seems like a necessary component (I actually that we already had it) if we are to use this as an interchange format, since republishing previously credited information without that credit seems like a bad idea.

validation/schema.json
```diff
@@ -218,6 +218,45 @@
...
+            "type": "array",
+            "items": {
+              "type": "object",
+              "properties": {
+                "type": {
+                  "type": "string",
+                  "enum": [
+                    "EMAIL",
+                    "GITHUB",
+                    "OTHER",
```
oliverchang, line 239: nit: could you please remove OTHER here as well, to match the documentation?

oliverchang (approved):
Thanks! Just one last comment on the validation.

rsc:
Overall, adding credit makes sense to me.

I wonder whether the enum of contact info makes sense, or if we should just make the contact list be a list of URLs. All the current ones would fit:

mailto:email@domain.com
https://twitter.com/username
https://github.com/username

and then we don't have to worry about maintaining an enumeration of social sites, nor defining "OTHER".

chrisbloom7:
@rsc Se updated spec

cc:/ @oliverchang since you seemed close to signing off on the previous version. Any objections to this format instead?

oliverchang (approved):
This looks great to me! Thanks @rsc for suggesting this simplication :)

PR #26, opened 2022-01-28
Add "Android" ecosystem.

baekhyunwook:
As a part of effort of exporting Android Security Bulletin (ASB) CVEs to OSV, we would need a new ecosystem for Android. This pull request is for updating the OSV schema description to add "Android" as a new ecosystem for ASB CVEs.

### DIFF
--- a/docs/schema.md
+++ b/docs/schema.md
@@ -1229,7 +1229,7 @@ incompatible changes will be made in any 1.x version.
 
 ## Android Ecosystem Components
 
-A component of `Android` ecosystem CVE indicates the component of Android firmware that the patch applies to. This matches to the component names used in [Android Security Bulletins (ASB)](https://source.android.com/security/bulletin). The complete list of components used in ASB is as follows:
+A package name in the `Android` ecosystem indicates the component of Android that the advisory applies to. This matches the component names used in [Android Security Bulletins (ASB)](https://source.android.com/security/bulletin). The complete list of components used in ASB is as follows:
 
 - Framework
 - Library
--- a/docs/schema.md
+++ b/docs/schema.md
@@ -376,6 +376,7 @@ The defined ecosystems are:
 | `Linux` | The Linux kernel. The only supported `name` is `Kernel`. |
 | `Debian` | The Debian package ecosystem; the `name` is the name of the package. |
 | `Hex` | The package manager for the Erlang ecosystem; the `name` is a Hex package name. |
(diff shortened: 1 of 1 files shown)


docs/schema.md
```diff
@@ -376,6 +376,7 @@ The defined ecosystems are:
 | `Linux` | The Linux kernel. The only supported `name` is `Kernel`. |
 | `Debian` | The Debian package ecosystem; the `name` is the name of the package. |
 | `Hex` | The package manager for the Erlang ecosystem; the `name` is a Hex package name. |
+| `Android`  | The Android ecosystem; the `name` field is the Android component name that the patch applies to, as shown in the [Android Security Bulletins](https://source.android.com/security/bulletin) such as `Framework`, `Media Framework` and `Kernel Component`. |
```
oliverchang, line 379: Is there an exhaustive list of these somewhere that you can reference here? Generally we try to ensure these values are consistent, precise and can be  programmatically matched to a component without any ambiguity (e.g. a package on "npm")

oliverchang (commented):
thanks!! just a question

docs/schema.md
baekhyunwook: Currently, there is no explicit list maintained on the public side of ASB,
but we manually queried to get the list and the result is as follows:

Qualcomm
Kernel
Media Framework
MediaTek
Framework
System
NVIDIA
HTC
LG
Android Runtime
Broadcom
Library
Marvell
FPC
AMLogic
Kernel LTS
ARM
Unisoc
Telecommunication
Widevine DRM
Android TV
Amlogic

Maintaining and publishing an exhaustive list on the public side of ASB is under discussion, but I think it won't be updated soon.  So, until it happens, as an alternative, I am considering instead adding the list to the description of `Android` ecosystem (and maintain). What do you think?

oliverchang: Thanks! These still seem very broad -- Is there something more fine grained here to identify packages within these things? 

If not, having the list maintained here for now SGTM! Since it's rather long, let's add an appendix section at the end and link to that from the main ecosystems list here.

baekhyunwook: Unfortunately, that is the finest granularity that ASB uses to distinguish the domain of a CVE.

Yes, adding them to an appendix will make it cleaner. I will make an update and send. Thanks!

baekhyunwook: Updated the commit to show the complete list of Android components at an Appendix section. Please take a look.

oliverchang (approved):
Thanks! I did some minor rewording of the appendix but this looks great!

PR #28, opened 2022-02-01
Update examples and add changelog for 1.2.0.

oliverchang:
Also clarify spec to state that the CVSS "score" is the vector string.

### DIFF
--- a/docs/schema.md
+++ b/docs/schema.md
@@ -272,7 +272,7 @@ describes the quantitative method used to calculate the associated `score`.
 
 | Severity Type | Score Description |
 | --------- | ----------- |
-| `CVSS_V3` | A computed score representing the unique characteristics and severity of the vulnerability using a version of the [Common Vulnerability Scoring System notation](https://www.first.org/cvss/) that is >= 3.0 and < 4.0. |
+| `CVSS_V3` | A CVSS vector string representing the unique characteristics and severity of the vulnerability using a version of the [Common Vulnerability Scoring System notation](https://www.first.org/cvss/) that is >= 3.0 and < 4.0 (e.g.`"CVSS:3.1/AV:N/AC:H/PR:N/UI:N/S:C/C:H/I:N/A:N"`).|
 | Your quantitative severity type here. | [Send us a PR](https://github.com/ossf/osv-schema/compare). |
 
 ### severity[].score field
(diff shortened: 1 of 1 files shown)


docs/schema.md
```diff
@@ -800,13 +802,27 @@ Neither GitHub nor NPM uses this format currently, but here is how a recent NPM
...
+    "schema_version": "1.2.0",
     "id": "GHSA-r9p9-mrjm-926w",
     "published": "2021-03-07T11:27:00Z",
     "modified": "2021-03-10T23:40:39Z",
     "aliases": ["NPM-1648", "CVE-2020-28498", "SNYK-JS-ELLIPTIC-1064899"],
     "related": ["NPM-1649", "SNYK-JAVA-ORGWEBJARSNPM-1069836"],
     "summary": "Use of a Broken or Risky Cryptographic Algorithm",
     "details": "elliptic is a Fast elliptic-curve cryptography in a plain javascript implementation.\n\nAffected versions of this package are vulnerable to Cryptographic Issues via the secp256k1 implementation in elliptic/ec/key.js. There is no check to confirm that the public key point passed into the derive function actually exists on the secp256k1 curve. This results in the potential for the private key used in this implementation to be revealed after a number of ECDH operations are performed.\n\nRemediation: Upgrade elliptic to version 6.5.4 or higher.\n",
+    "severity": {
+        {"type": "CVSS_V3", "score": "6.8"}
```
chrisbloom7, line 814: Our intention for this was that the `score` would be the CVSS v3 vector string, for example `"CVSS:3.1/AV:N/AC:H/PR:N/UI:N/S:C/C:H/I:N/A:N"` (what you had labeled `code` in the old `database_specific` example), not the numeric score. This may have been a miscommunication on my part. Do you feel like this field as defined will work for that, or do we need to clarify it's intention, or perhaps we need to revisit a top level CVSS field if we want to be able to communicate several aspects of the CVSS calculations.

oliverchang: Ah, I see! Sorry for the misunderstanding. In that case, let's clarify the spec to say that this value has to be the vector string. I'll do that in this PR as well!

oliverchang: @chrisbloom7 done!

chrisbloom7 (approved):
Looks great!

Issue #27, opened 2022-01-31
affected version assumptions/logic clarification


kurtseifried:
So I read https://ossf.github.io/osv-schema/ and I think there's an underlying assumption that should maybe be made explicit:

In most cases, there should be exactly one entry in the affected array per affected package to describe all affected versions. In rare cases, for example if the ecosystem_specific encodes platform information that doesn’t apply equally to all listed versions and ranges, a separate entry with the same package in the affected array may be needed.

The versions field can enumerate a specific set of affected versions, and the ranges field can list ranges of affected versions, under a given defined ordering. A version is considered affected if it lies within any one of the ranges or is listed in the versions list.

So in theory (assuming the data is correct) anything NOT listed is explicitly NOT vulnerable, correct? Should this logic be made explicit (assuming it is correct)? 

I also wonder if it might be helpful to also support an "unknown" or "needs investigation:" type tag, this would for example allow people to publish OSV's quicker ("we know this specific version is affected, but we haven't looked at much older code yet, someone should probably do that") with a finer degree of detail, or is this overengineering and not needed (I'm leaning in this direction mostly right now).

oliverchang:
Yep! Anything not listed is explicitly not vulnerable. Our rationale was that any uncertain versions are not very actionable for consumers and vulnerability scanners, so we would lean against trying to encode this complexity into the schema. 

We could add a sentence along the lines of "Any versions not covered is assumed to be unaffected." after "A version is considered affected if it lies within any one of the ranges or is listed in the versions list." to make this more explicit? 

Our pseudocode for evaluation also makes this clear:  https://ossf.github.io/osv-schema/#evaluation

kurtseifried:
Sounds good, also it may be worth mentioning the best way to mark software that is not vulnerable if there are none, e.g. use "*" vs. marking "everything up until now" and when it may be appropriate to use *, e.g. I'm thinking cases like Red Hat Satellite 6 and its use of Jackson Databind, basically it didn't use the deserialization so most of the really bad CVEs back in the day didn't affect it, provably so.

oliverchang:
Thinking more it seems better to improve the psuedocode and link to it from the text you pointed out to remove any ambiguity. I did this here: https://github.com/ossf/osv-schema/pull/29

> best way to mark software that is not vulnerable if there are none, e.g. use "*" vs. marking "everything up until now" and when it may be appropriate to use 

[...]

PR #29, opened 2022-02-02
Improve pseudocode for evaluation.

oliverchang:
Improve the pseudocode to cover matching `affected.versions` and
`affected.package` as well.

Link to this from earlier in the spec.

Clearing up some ambiguity pointed out in #27


inferno-chromium (approved):

chrisbloom7 (approved):
🎉

Issue #35, opened 2022-03-03
Enable GHSA to consolidate `affected` entries


oliverchang:
## Context
Currently, GitHub's Advisory Database encodes pairs of (introduced, fixed) events in separate `affected` objects. This is because GitHub needs to preserve these pairs when converting the OSV entry back into their internal database represenation. 

An example: https://github.com/github/advisory-database/blob/d6004eb8de91ad341605da869ab1b9f1e4abe433/advisories/github-reviewed/2017/10/GHSA-xgr2-v94m-rc9g/GHSA-xgr2-v94m-rc9g.json#L14

```json
"affected": [
    {
      "package": {
        "ecosystem": "RubyGems",
        "name": "activesupport"
      },
      "ranges": [
        {
          "type": "ECOSYSTEM",
          "events": [
            { "introduced": "2.3.2" },
            { "fixed": "2.3.16" }
          ]
        }
      ]
    },
    {
      "package": {
        "ecosystem": "RubyGems",
        "name": "activesupport"
      },
      "ranges": [
        {
          "type": "ECOSYSTEM",
          "events": [
            { "introduced": "3.0.0" },
            { "fixed": "3.0.20" }
          ]
        }
      ]
    }
  ],
```

One obvious alternative here is to use a single `affected` entry and have two separate `ECOSYSTEM` ranges in there:

```json
      "ranges": [
        {
          "type": "ECOSYSTEM",
          "events": [
            { "introduced": "2.3.2" },
            { "fixed": "2.3.16" }
          ]
        }, 
        {
          "type": "ECOSYSTEM",
          "events": [
            { "introduced": "3.0.0" },
            { "fixed": "3.0.20" }
          ]
        }
      ]
```

However, I believe this wasn't done, because GitHub needs to record version metadata using a <= operator, and relies on `database_specific` to this. One example: https://github.com/github/advisory-database/blob/d6004eb8de91ad341605da869ab1b9f1e4abe433/advisories/github-reviewed/2020/03/GHSA-gww7-p5w4-wrfv/GHSA-gww7-p5w4-wrfv.json 

This has a bunch of affected entries that look like:

### GHSA-gww7-p5w4-wrfv.json
```json
    {
      "ranges": [
        {
          "type": "ECOSYSTEM",
          "events": [
            {
              "introduced": "2.0.0"
            }
          ]
        }
      ],
      "database_specific": {
        "last_known_affected_version_range": "<= 2.0.6"
      }
    },
```

## Problems

[...]

PR #38, opened 2022-03-17
Add `last_affected` event type.

oliverchang:
Part of #35.

### DIFF
--- a/docs/schema.md
+++ b/docs/schema.md
@@ -444,7 +444,7 @@ The `ranges` object's `events` field is a JSON array of objects. Each object
 describes a single version that either:
 1. Introduces a vulnerability: `{"introduced": string}`
 2. Fixes a vulnerability: `{"fixed": string}`
-3. Describes the last affected version: `{"last_affected": string}`
+3. Describes the last known affected version: `{"last_affected": string}`
 4. Sets an upper limit on the range being described: `{"limit": string}`
 
 These `events` objects represent a "timeline" of status changes for the affected
--- a/docs/schema.md
+++ b/docs/schema.md
@@ -81,6 +81,7 @@ A JSON Schema for validation is also available
 			"events": [ {
 				"introduced": string,
 				"fixed": string,
+				"last_affected": string,
 				"limit": string
 			} ]
 		} ],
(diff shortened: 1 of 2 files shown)


docs/schema.md
```diff
@@ -443,7 +444,8 @@ The `ranges` object's `events` field is a JSON array of objects. Each object
 describes a single version that either:
 1. Introduces a vulnerability: `{"introduced": string}`
 2. Fixes a vulnerability: `{"fixed": string}`
-3. Sets an upper limit on the range being described: `{"limit": string}`
+3. Describes the last affected version: `{"last_affected": string}`
```
chrisbloom7, line 447:
```suggestion
3. Describes the last known affected version: `{"last_affected": string}`
```

validation/schema.json
```diff
@@ -132,6 +132,17 @@
...
                       },
+                      {
+                        "type": "object",
+                        "properties": {
+                          "last_affected": {
+                            "type": "string"
+                          }
+                        },
+                        "required": [
+                          "last_affected"
```
chrisbloom7, line 143: Should we also add validation to assert that a `fixed` event cannot occur alongside a `last_effected` event and vice versa?

chrisbloom7 (approved):
LGTM! I left a couple suggestions that aren't blockers

validation/schema.json
oliverchang: Thanks, that's a good idea! I'll look at doing this.

That said, I'm wondering how this would look like for GHSA specifically when an entry wants to encode both <= and < (the patched version)? Would you just be using "fixed" here to capture the < and then database_specific to capture the <= ?

chrisbloom7: Currently if an advisory has a fix we replace the upper bound with `< f.i.x` when we rehydrate and consider it an acceptable change since. It's only for advisories that don't have a fix that we need to add the rehydration hints.

chrisbloom7: And same thing for the other direction when transforming to OSV - if an advisory has a fix, we add the `fixed` event and don't add the rehydration hint.

oliverchang: Thanks for clarifying!

oliverchang: Also added the JSON validation.

rsc (approved):
Looks great, thanks.

PR #43, opened 2022-04-21
feat: add LoopBack Advisory Database

achrinza:
This PR adds reference to the LoopBack Advisory Database.

## Background

As part of the ongoing security-related enhancement works, LoopBack is currently setting up its own advisory database for Node.js modules under its purview.

We're working to republish existing advisories under machine-readable formats (which includes OSV) and with the `LBSEC-` prefix.

see: https://github.com/loopbackio/security/issues/14

### DIFF
--- a/docs/schema.md
+++ b/docs/schema.md
@@ -158,6 +158,7 @@ The defined database prefixes and their "home" databases are:
 | `RUSTSEC` | [The Rust crates vulnerability database](https://github.com/rustsec/advisory-db). Serving `<ID>` in the shared format at  `https://github.com/RustSec/advisory-db/blob/osv/crates/<ID>.json` |
 | `GSD` | The GSD database. Serving the shared format [here](https://github.com/cloudsecurityalliance/gsd-database). |
 | `GHSA` | The GitHub Security Advisory database. Serving the shared format [here](https://github.com/github/advisory-database). |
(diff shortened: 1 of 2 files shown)


docs/schema.md
```diff
@@ -158,6 +158,7 @@ The defined database prefixes and their "home" databases are:
 | `RUSTSEC` | [The Rust crates vulnerability database](https://github.com/rustsec/advisory-db). Serving `<ID>` in the shared format at  `https://github.com/RustSec/advisory-db/blob/osv/crates/<ID>.json` |
 | `GSD` | The GSD database. Serving the shared format [here](https://github.com/cloudsecurityalliance/gsd-database). |
 | `GHSA` | The GitHub Security Advisory database. Serving the shared format [here](https://github.com/github/advisory-database). |
+| `LBSEC` | The LoopBack Advisory Database. Serving the shared format [here](https://github.com/loopbackio/security/tree/main/advisories). |
```
chrisbloom7, line 161: The examples in the accompanying repo indicate these will use an `LBSA` prefix, not `LBSEC` - is this still in flux?

achrinza: `LBSA-` was used interim when we were initially setting up the repo but we've since switched over to `LBSEC-`, which would be reflected in a PR that we're planning to land in the upcoming week or so.

chrisbloom7: @achrinza thanks for clarifying!

oliverchang (approved):
Thanks!

PR #46, opened 2022-06-03
Script for converting debian security tracker to osv format 

another-rex:
Also adds a markdownify dependency

### DIFF
--- a/tools/debian/convert_debian.py
+++ b/tools/debian/convert_debian.py
@@ -253,9 +253,7 @@ def convert_debian(webwml_repo: str, security_tracker_repo: str,
     advisories: Advisories = {}
 
     parse_security_tracker_file(advisories, security_tracker_repo)
-
     parse_webwml_files(advisories, webwml_repo)
-
     write_output(output_dir, advisories)
 
 
--- a/tools/debian/convert_debian.py
+++ b/tools/debian/convert_debian.py
@@ -39,7 +39,7 @@ VERSION_PATTERN = re.compile(r'\[(.*?)\]\s*-\s*([^\s]+)\s*([^\s]+)')
 WML_DESCRIPTION_PATTERN = re.compile(
     r'<define-tag moreinfo>((?:.|\n)*)</define-tag>', re.MULTILINE)
 
-# e.g. <define-tag report_date>(.*)</define-tag>
+# e.g. <define-tag report_date>2022-1-04</define-tag>
 WML_REPORT_DATE_PATTERN = re.compile(
     r'<define-tag report_date>(.*)</define-tag>')
 
@@ -187,18 +187,16 @@ def parse_webwml_files(advisories: Advisories, webwml_repo: str):
         val_wml = file_path_map.get(mapped_key_no_ext + '.wml')
         val_data = file_path_map.get(mapped_key_no_ext + '.data')
 
-        if not val_wml:
-            print(mapped_key_no_ext)
-        else:
+        if val_wml:
             with open(val_wml, encoding='utf-8') as handle:
                 data = handle.read()
                 html = WML_DESCRIPTION_PATTERN.findall(data)[0]
                 res = markdownify.markdownify(html)
                 adv.details = res
-
-        if not val_data:
-            print(mapped_key_no_ext)
         else:
+            print('No WML file yet for this:' + mapped_key_no_ext)
+
+        if val_data:
             with open(val_data, encoding='utf-8') as handle:
                 data: str = handle.read()
                 report_date: str = WML_REPORT_DATE_PATTERN.findall(data)[0]
--- a/tools/debian/convert_debian.py
+++ b/tools/debian/convert_debian.py
@@ -16,14 +16,13 @@ import argparse
 import json
 import os
 import re
-import traceback
-from typing import Any, Dict, Optional, TextIO
-import xml.etree.ElementTree as ET
-import markdownify
+from typing import TextIO
 import datetime
 
-import osv
-import osv.ecosystems
+import markdownify
+
+# import osv
+# import osv.ecosystems
 
 WEBWML_SECURITY_PATH = os.path.join('english', 'security')
 SECURITY_TRACKER_DSA_PATH = os.path.join('data', 'DSA', 'list')
@@ -37,10 +36,15 @@ DSA_PATTERN = re.compile(r'\[.*?\]\s*([\w-]+)\s*(.*)')
 VERSION_PATTERN = re.compile(r'\[(.*?)\]\s*-\s*([^\s]+)\s*([^\s]+)')
 
 # e.g. <define-tag moreinfo>\n Some html here \n</define-tag>
-WML_DESCRIPTION_PATTERN = re.compile(r'<define-tag moreinfo>((?:.|\n)*)</define-tag>', re.MULTILINE)
+WML_DESCRIPTION_PATTERN = re.compile(
+    r'<define-tag moreinfo>((?:.|\n)*)</define-tag>', re.MULTILINE)
 
-# e.g.
-WML_REPORT_DATE_PATTERN = re.compile(r'<define-tag report_date>(.*)</define-tag>')
+# e.g. <define-tag report_date>(.*)</define-tag>
+WML_REPORT_DATE_PATTERN = re.compile(
+    r'<define-tag report_date>(.*)</define-tag>')
+
+# e.g. DSA-12345-2, -2 is the extension
+MATCH_EXTENSION_FROM_DSA = re.compile(r'-\d+$')
 
 
 class DebianSpecificInfo:
@@ -55,6 +59,7 @@ class DebianSpecificInfo:
 
 class AffectedInfo:
     """Debian version info."""
+
     ecosystem_specific: DebianSpecificInfo
     package: str
     ranges: [str]
@@ -66,21 +71,20 @@ class AffectedInfo:
         self.package = package
         self.fixed = fixed
 
-    def toJSON(self):
+    def to_json(self):
         return {
-            "ecosystem_specific": self.ecosystem_specific,
-            "package": {
-                "ecosystem": "Debian",
-                "name": self.package
+            'ecosystem_specific':
+                self.ecosystem_specific,
+            'package': {
+                'ecosystem': 'Debian',
+                'name': self.package
             },
-            "ranges": [
-                {
-                    "type": "ECOSYSTEM",
-                    "events": [
-                        {"fixed": self.fixed}
-                    ]
-                }
-            ],
+            'ranges': [{
+                'type': 'ECOSYSTEM',
+                'events': [{
+                    'fixed': self.fixed
+                }]
+            }],
         }
 
     def __repr__(self):
@@ -89,6 +93,7 @@ class AffectedInfo:
 
 class AdvisoryInfo:
     """Debian advisory info."""
+
     id: str
     summary: str
     details: str
@@ -96,26 +101,31 @@ class AdvisoryInfo:
     affected: [AffectedInfo]
     aliases: [str]
 
-    def __init__(self, id, summary):
-        self.id = id
+    def __init__(self, adv_id, summary):
+        self.id = adv_id
         self.summary = summary
         self.affected = []
         self.aliases = []
-        self.published = ""
-        self.details = ""
+        self.published = ''
+        self.details = ''
 
     def __repr__(self):
         return json.dumps(self, default=dumper, indent=2)
 
 
+Advisories = dict[str, AdvisoryInfo]
+"""Type alias for collection of advisory info"""
+
+
 def dumper(obj):
     try:
-        return obj.toJSON()
+        return obj.to_json()
     except AttributeError:
         return obj.__dict__
 
 
-def parse_security_tracker_file(advisories: dict[str, AdvisoryInfo], file_handle: TextIO):
+def parse_security_tracker_file(advisories: Advisories,
+                                file_handle: TextIO):
     current_advisory = None
 
     for line in file_handle:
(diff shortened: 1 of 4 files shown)


tools/debian/convert_debian.py
```diff
@@ -34,97 +36,179 @@
 # e.g. [buster] - xz-utils 5.2.4-1+deb10u1
 VERSION_PATTERN = re.compile(r'\[(.*?)\]\s*-\s*([^\s]+)\s*([^\s]+)')
 
+# e.g. <define-tag moreinfo>\n Some html here \n</define-tag>
+WML_DESCRIPTION_PATTERN = re.compile(r'<define-tag moreinfo>((?:.|\n)*)</define-tag>', re.MULTILINE)
```
oliverchang, line 40: Can you make all lines <= 80 columns? 

You can use `yapf` and `pylint` to fix/find these issues.

tools/debian/convert_debian.py
```diff
@@ -34,97 +36,179 @@
...
 
-class VersionInfo:
-    """Debian version info."""
+# e.g.
+WML_REPORT_DATE_PATTERN = re.compile(r'<define-tag report_date>(.*)</define-tag>')
+
+
+class DebianSpecificInfo:
```
oliverchang, line 46: Please add a docstring to describe what this class does.

tools/debian/convert_debian.py
```diff
@@ -34,97 +36,179 @@
...
 
     def __init__(self, release: str, package: str, fixed: str):
-        self.release = release
+        self.ecosystem_specific = DebianSpecificInfo(release)
         self.package = package
         self.fixed = fixed
 
+    def toJSON(self):
+        return {
+            "ecosystem_specific": self.ecosystem_specific,
```
oliverchang, line 71: Please use single quotes for all strings (except for """ docstrings).

tools/debian/convert_debian.py
```diff
@@ -34,97 +36,179 @@
...
+
+    file_path_map = {}
+
+    for root, dirs, files in os.walk(os.path.join(webwml_repo, WEBWML_SECURITY_PATH)):
+        for file in files:
+            file_path_map[file] = os.path.join(root, file)
+
+    # Add descriptions to advisories from wml files
+    for key in advisories:
+        # TODO: use regex to remove suffix?
```
oliverchang, line 187: +1. Please use a regex to replace these.

tools/debian/convert_debian.py
```diff
@@ -34,97 +36,179 @@
...
+        mapped_key_no_ext = key.lower().removesuffix('-1').removesuffix('-2').removesuffix('-3').removesuffix('-4')
+        val_wml = file_path_map.get(mapped_key_no_ext + ".wml")
+        val_data = file_path_map.get(mapped_key_no_ext + ".data")
+
+        if not val_wml:
+            print(mapped_key_no_ext)
```
oliverchang, line 193: Is this debugging code? can remove if so.

tools/debian/convert_debian.py
```diff
@@ -34,97 +36,179 @@
...
+                res = markdownify.markdownify(html)
+                advisories[key].details = res
+
+        if not val_data:
+            print(mapped_key_no_ext)
+        else:
+            with open(val_data) as handle:
+                data: str = handle.read()
+                report_date: str = WML_REPORT_DATE_PATTERN.findall(data)[0]
+                # Split by ',' here for the occasional case where there are two dates in the "publish" field
```
oliverchang, line 207: Can you link to some examples? Curious which one is the right one.

tools/debian/convert_debian.py
```diff
@@ -34,97 +36,179 @@
...
-                current_advisory = dsa_match.group(1)
-                advisories[current_advisory] = AdvisoryInfo(
-                    current_advisory, dsa_match.group(2))
-
-    print(advisories)
+        parse_security_tracker_file(advisories, handle)
+
+    file_path_map = {}
+
+    for root, dirs, files in os.walk(os.path.join(webwml_repo, WEBWML_SECURITY_PATH)):
```
oliverchang, line 181: nit: use '_' instead of 'dirs'.

tools/debian/convert_debian.py
oliverchang, line 40: Is there a more robust/reliable way to parse these? e.g. using an xml parser?

tools/debian/convert_debian.py
```diff
@@ -34,97 +36,179 @@
...
     fixed: str
+    versions: [str]
 
     def __init__(self, release: str, package: str, fixed: str):
-        self.release = release
+        self.ecosystem_specific = DebianSpecificInfo(release)
         self.package = package
         self.fixed = fixed
 
+    def toJSON(self):
```
oliverchang, line 69: please use snake_case for all function names.

Also, can we just use `__dict__` for this? That's more accurately describes what this is doing and makes the `dumper` below much simpler.

oliverchang (commented):
very nice work!! some initial comments.

Please see https://google.github.io/styleguide/pyguide.html for our style guide as well.

tools/debian/convert_debian.py
another-rex: Turns out there was more than I originally thought.

Most of them is from 2006 - 2008, with the exception the one in 2013. The 2013 one has a extra dash after the date and no space between the comma and date, perhaps it's copied / pasted incorrectly.

```
2013-8-23,2013-9-01-
DSA-2740-2
2013-8-23,2013-9-01-
DSA-2740-1
2006-9-25, 2006-9-26
DSA-1184-2
2005-10-1, 2005-10-4
DSA-833-2
2008-2-11, 2008-2-12
DSA-1494-1
2007-10-2, 2007-10-12
DSA-1381-2
2007-3-20, 2007-03-28
DSA-1270-1
2006-12-3, 2006-12-3
DSA-1225-1
2006-11-30, 2006-12-1
DSA-1222-1
2006-11-20, 2006-12-27
DSA-1214
2006-11-12, 2006-11-13
DSA-1209
2006-11-9, 2006-11-19
DSA-1207-1
2006-11-2, 2006-12-1
DSA-1205-1
2006-9-28, 2006-10-2
DSA-1185-2
2006-9-3, 2006-10-13
DSA-1166
2006-8-29, 2006-9-13
DSA-1161
```

tools/debian/convert_debian.py
another-rex: The .wml and .data files seem to take a bit of time to generate and get placed in the repository. So for the latest few DSA's the corresponding .wml file will not exist. I'm not sure what the desired behavior should be for this, either:

- Just generate a partial .json file
- Not generate anything for now and skip the latest DSA's

For now I updated it to emit a descriptive message about the wml file not existing.

tools/debian/convert_debian.py
another-rex: thought of using an XML parser, but the default python one doesn't seem to like the format of the `.wml` and `.data` files. 
I'll look into this more, maybe it's a configuration I need to set.

oliverchang: Ack. Can you add a TODO here?

tools/debian/convert_debian.py
oliverchang: This is resolved.

tools/debian/convert_debian.py
```diff
@@ -34,97 +35,204 @@
...
+
+def dumper(obj):
+    try:
+        return obj.to_json()
+    except AttributeError:
+        return obj.__dict__
+
+
+def parse_security_tracker_file(advisories: Advisories,
+                                file_handle: TextIO):
```
oliverchang, line 131: Please add a docstring.

tools/debian/convert_debian.py
```diff
@@ -34,97 +35,204 @@
...
     fixed: str
+    versions: [str]
 
     def __init__(self, release: str, package: str, fixed: str):
-        self.release = release
+        self.ecosystem_specific = DebianSpecificInfo(release)
         self.package = package
         self.fixed = fixed
 
+    def to_json(self):
```
oliverchang, line 77: This isn't really returning JSON right? Let's rename this to `to_dict`.

tools/debian/convert_debian.py
```diff
@@ -34,97 +35,211 @@
...
+        if val_wml:
+            with open(val_wml, encoding='utf-8') as handle:
+                data = handle.read()
+                html = WML_DESCRIPTION_PATTERN.findall(data)[0]
+                res = markdownify.markdownify(html)
+                adv.details = res
+        else:
+            print('No WML file yet for this: ' + mapped_key_no_ext)
+
+        if val_data:
```
oliverchang, line 202: nit: prefer

```python
if not val_data:
  continue
```

to reduce indentation.

tools/debian/convert_debian.py
```diff
@@ -34,97 +35,211 @@
...
+def parse_webwml_files(advisories: Advisories, webwml_repo: str):
+    file_path_map = {}
+
+    for root, _, files in os.walk(
+            os.path.join(webwml_repo, WEBWML_SECURITY_PATH)):
+        for file in files:
+            file_path_map[file] = os.path.join(root, file)
+
+    # Add descriptions to advisories from wml files
+    for key, adv in advisories.items():
```
oliverchang, line 186: let's be more descriptive than "key" and "adv".

We can do:

```python
for dsa_id, advisory in advisories.items():
  ...
```

tools/debian/convert_debian.py
```diff
@@ -34,97 +35,211 @@
...
+                #     print(key)
+
+                # Split by ',' here for the occasional case where there
+                # are two dates in the 'publish' field
+                adv.published = (datetime.datetime.strptime(
+                    report_date.split(',')[0], '%Y-%m-%d').isoformat() + 'Z')
+
+
+def write_output(output_dir: str, advisories: Advisories):
+    for key, value in advisories.items():
```
oliverchang, line 219: same here: let's use "dsa_id" and "advisory" respectively for these varibles.

tools/debian/convert_debian.py
```diff
@@ -89,33 +93,39 @@ def __repr__(self):
...
+        self.id = adv_id
         self.summary = summary
         self.affected = []
         self.aliases = []
-        self.published = ""
-        self.details = ""
+        self.published = ''
+        self.details = ''
 
     def __repr__(self):
```
oliverchang, line 115: Let's be explicit and add a `to_dict` method to all of these custom classes. the `__repr__` function can just call that.

tools/debian/convert_debian.py
```diff
@@ -154,59 +167,70 @@ def parse_security_tracker_file(advisories: dict[str, AdvisoryInfo], file_handle
...
+                adv.published = (datetime.datetime.strptime(
+                    report_date.split(',')[0], '%Y-%m-%d').isoformat() + 'Z')
+
+
+def write_output(output_dir: str, advisories: Advisories):
+    for key in advisories:
+        with open(os.path.join(output_dir, key + '.json'),
+                  'w',
+                  encoding='utf-8') as output_file:
+            output_file.write(str(advisories[key]))
```
oliverchang: Let's be explicit here and call json.dumps() here on `advisory.to_dict()`. This makes it clear to the reader what's going on here.

tools/debian/convert_debian.py
```diff
@@ -154,59 +167,70 @@ def parse_security_tracker_file(advisories: dict[str, AdvisoryInfo], file_handle
...
-                # Split by ',' here for the occasional case where there are two dates in the "publish" field
-                advisories[key].published = \
-                    datetime.datetime.strptime(report_date.split(",")[0], "%Y-%m-%d").isoformat() + "Z"
+                # Split by ',' here for the occasional case where there
+                # are two dates in the 'publish' field
+                adv.published = (datetime.datetime.strptime(
+                    report_date.split(',')[0], '%Y-%m-%d').isoformat() + 'Z')
+
+
+def write_output(output_dir: str, advisories: Advisories):
```
oliverchang, line 218: Please add a docstring.

tools/debian/convert_debian.py
```diff
@@ -154,59 +167,70 @@ def parse_security_tracker_file(advisories: dict[str, AdvisoryInfo], file_handle
...
+    for key in advisories:
```
oliverchang: Use advisories.items() for iterating here.

oliverchang (commented):
Nice! This looks good with some remaining nits/comments.

tools/debian/convert_debian.py
another-rex: Did some more testing here, turns out the `.data` format is not valid XML, since all attributes in XML has to have a value associated with it (at least according to this answer: https://stackoverflow.com/a/6928130). 
 
I tried a html parser instead and it sort of works, though required wrapping the file with a root element. I'm not sure if that's much more reliable compared to the regex version. 

I have just added a TODO there for now.

tools/debian/convert_debian.py
oliverchang: I think the first date is the correct publihsed date. 

It looks like significant revisions result in a new date being added: https://salsa.debian.org/webmaster-team/webwml/-/commit/e1fa82c1bc3e984b17a65d980ffbb38b49f21087

tools/debian/convert_debian.py
```diff
@@ -34,58 +34,118 @@
...
-            'release': self.release,
-            'package': self.package,
-            'fixed': self.fixed,
-        })
+        return json.dumps(self, default=dumper)
 
 
 class AdvisoryInfo:
     """Debian advisory info."""
+
```
oliverchang, line 105: Can you add a TODO for adding `modified` as well? We can probably take this from the git log.

tools/debian/convert_debian.py
oliverchang: Ack, regex may be OK in this case then. Thanks -- TODO sounds good!

tools/debian/convert_debian.py
oliverchang: Let's generate a partial .json file, and fill the description/summary with boilerplate (e.g. "Details not yet available").

another-rex: Added a TODO there for now

tools/debian/convert_debian.py
```diff
@@ -124,7 +189,74 @@ def convert_debian(webwml_repo: str, security_tracker_repo: str,
...
+    print('Complete')
+
+
+def convert_debian(webwml_repo: str, security_tracker_repo: str,
+                   output_dir: str):
+    """Convert Debian advisory data into OSV."""
+    advisories: Advisories = {}
+
+    parse_security_tracker_file(advisories, security_tracker_repo)
+
```
oliverchang, line 256: nit: no need for these newlines on line 256 and 258.

oliverchang (approved):
LGTM with one last nit!

Issue #24, opened 2022-01-25
Support Debian advisories


oliverchang:
Moving discussion from https://github.com/ossf/osv-schema/pull/10.

oliverchang:
@rhertzog @jeremiah @pombredanne Moving our discussion from https://github.com/ossf/osv-schema/pull/10 and coming back on this thread once more. I really apologize for the delay -- I really need to fix my GitHub email filters. 

To answer some questions from the last comment:

> What kind of generic tools are you referring too? How would they consume this data?

These would be SBOM (e.g. CycloneDX, SPDX) generator tools (which e.g. generate this from a container image) and generic vulnerability scanner / matchers which take an SBOM or similar as input. Typically these work at a binary package level so having this list of binary packages will make things much easier. Our goal with the OSV schema is to define a generic format for describing vulnerabilities that plugs into generic tooling like these without users having to understand the details of the ecosystem.

The size of the submitted data will increase, yes, but this shouldn't be a problem as they're being consumed by automated scanners and vulnerability matchers. 

Are the existing [Debian advisories](https://www.debian.org/security/2021/) available in a machine parseable format anywhere? We'd be happy to convert a few entries manually to show what these would look like, and  help build tooling to expand / convert these from existing DSA entries.

rhertzog:
> Typically these work at a binary package level so having this list of binary packages will make things much easier. 

If they do get list of Debian binary packages, they already access the dpkg database and in that case, it's just as convenient to extract the source package name from the database. It's just another field in the same file. Or another query to dpkg-query.

We should not make it convenient for tools that consume the data, but for humans that build and maintain the database. tools can be improved and will not complain about the extra step.

> Are the existing Debian advisories available in a machine parseable format anywhere?
Our security data is managed in the security tracker: https://security-tracker.debian.org/tracker/
Most of the data is in text files in a git repository: https://salsa.debian.org/security-tracker-team/security-tracker/-/blob/master/data/DSA/list

But we have some JSON export built by the security tracker: https://security-tracker.debian.org/tracker/data/json (beware it's big)

oliverchang:
> If they do get list of Debian binary packages, they already access the dpkg database and in that case, it's just as convenient to extract the source package name from the database. It's just another field in the same file. Or another query to dpkg-query.

[...]

PR #49, opened 2022-06-15
Improved Debian Importer

another-rex:
Adds the ability to have introduced version per Debian release. 
Adds method for git to retrieve last modified time (Commented out at the moment)
Formatted as Option 2 described in #24

### DIFF
--- a/tools/debian/convert_debian.py
+++ b/tools/debian/convert_debian.py
@@ -56,100 +56,104 @@ CAPTURE_DSA_WITH_NO_EXT = re.compile(r'dsa-\d+')
 
 
 class DebianSpecificInfo:
-    """Specific Debian information, exports to the
-    `ecosystem_specific` field in osv"""
+  """Specific Debian information, exports to the
 
-    # Debian release name
-    release: str
+  `ecosystem_specific` field in osv
+  """
 
-    # Debian release version number
-    version: str
+  # Debian release name
+  release: str
 
-    def __init__(self, release: str, version: str):
-        self.release = release
-        self.version = version
+  # Debian release version number
+  version: str
 
-    # try:
-    #         self.version = codename_to_version[self.release]
-    #     except KeyError as e:
-    #         raise Exception("Debian release name does not exist in map") from e
+  def __init__(self, release: str, version: str):
+    self.release = release
+    self.version = version
 
-    def to_dict(self):
-        return self.__dict__
+  # try:
+  #         self.version = codename_to_version[self.release]
+  #     except KeyError as e:
+  #         raise Exception("Debian release name does not exist in map") from e
 
-    def __repr__(self):
-        return json.dumps(self, default=dumper)
+  def to_dict(self):
+    return self.__dict__
+
+  def __repr__(self):
+    return json.dumps(self, default=dumper)
 
 
 class AffectedInfo:
-    """Debian version info."""
+  """Debian version info."""
 
-    ecosystem_specific: DebianSpecificInfo
-    package: str
-    ranges: [str]
-    fixed: str
-    introduced: str
-    versions: [str]
+  ecosystem_specific: DebianSpecificInfo
+  package: str
+  ranges: [str]
+  fixed: str
+  introduced: str
+  versions: [str]
 
-    def __init__(self, release: str, version: str, package: str, fixed: str, introduced: str):
-        self.ecosystem_specific = DebianSpecificInfo(release, version)
-        self.package = package
-        self.fixed = fixed
-        self.introduced = introduced
+  def __init__(self, release: str, version: str, package: str, fixed: str,
+               introduced: str):
+    self.ecosystem_specific = DebianSpecificInfo(release, version)
+    self.package = package
+    self.fixed = fixed
+    self.introduced = introduced
 
-    def to_dict(self):
-        return {
-            'ecosystem_specific':
-                self.ecosystem_specific,
-            'package': {
-                'ecosystem': 'debian:' + self.ecosystem_specific.version,
-                'name': self.package
-            },
-            'ranges': [{
-                'type': 'ECOSYSTEM',
-                'events': [
-                    {'introduced': self.introduced},
-                    {'fixed': self.fixed}
-                ]
-            }],
-        }
+  def to_dict(self):
+    return {
+        'ecosystem_specific':
+            self.ecosystem_specific,
+        'package': {
+            'ecosystem': 'debian:' + self.ecosystem_specific.version,
+            'name': self.package
+        },
+        'ranges': [{
+            'type': 'ECOSYSTEM',
+            'events': [{
+                'introduced': self.introduced
+            }, {
+                'fixed': self.fixed
+            }]
+        }],
+    }
 
-    def __repr__(self):
-        return json.dumps(self, default=dumper)
+  def __repr__(self):
+    return json.dumps(self, default=dumper)
 
 
 class AdvisoryInfo:
-    """Debian advisory info."""
+  """Debian advisory info."""
 
-    id: str
-    summary: str
-    details: str
-    published: str
-    modified: str
-    affected: [AffectedInfo]
-    aliases: [str]
-    # Internal use for the script
-    preexisting: bool
+  id: str
+  summary: str
+  details: str
+  published: str
+  modified: str
+  affected: [AffectedInfo]
+  aliases: [str]
+  # Internal use for the script
+  preexisting: bool
 
-    def __init__(self, adv_id, summary):
-        self.id = adv_id
-        self.summary = summary
-        self.affected = []
-        self.aliases = []
-        self.published = ''
-        self.details = ''
-        self.preexisting = False
+  def __init__(self, adv_id, summary):
+    self.id = adv_id
+    self.summary = summary
+    self.affected = []
+    self.aliases = []
+    self.published = ''
+    self.details = ''
+    self.preexisting = False
 
-    def to_dict(self):
-        result = copy.copy(self.__dict__)
-        # Remove the preexisting key from the json output since
-        # it's only for internal use in the script
+  def to_dict(self):
+    result = copy.copy(self.__dict__)
+    # Remove the preexisting key from the json output since
+    # it's only for internal use in the script
 
-        result.pop('preexisting')
-        return result
+    result.pop('preexisting')
+    return result
 
-    def __repr__(self):
-        return json.dumps(self, default=dumper)
+  def __repr__(self):
+    return json.dumps(self, default=dumper)
 
 
 Advisories = dict[str, AdvisoryInfo]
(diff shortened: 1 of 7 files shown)


tools/debian/first_package_finder.py
```diff
@@ -0,0 +1,120 @@
...
+  # `.gz` format always exist for all snapshots
+  source_path = '{version}/main/source/Sources.gz'
+
+  formatted_date = convert_datetime_to_str_datetime(date)
+
+  if version is None:
+    # return url for retrieving all versions for date
+    return base_url.format(date=formatted_date)
+  else:
+    return base_url.format(date=formatted_date) \
```
oliverchang, line 22: nit: no need for an else after return.

tools/debian/first_package_finder.py
```diff
@@ -0,0 +1,120 @@
...
+
+def convert_datetime_to_str_datetime(input_datetime: datetime) -> str:
+  """Convert datetime object to debian snapshot url string"""
+  return input_datetime.isoformat().replace('-', '').replace(':', '') + 'Z'
+
+
+def create_codename_to_version() -> pd.DataFrame:
+  """Returns the codename to version mapping"""
+  with request.urlopen(
+      'https://debian.pages.debian.net/distro-info-data/debian.csv') as csv:
```
oliverchang, line 34: nit: put this string in a global CONSTANT

tools/debian/convert_debian.py
```diff
@@ -51,227 +56,307 @@
...
+      # SimpleNamespace loads in the json as a
+      # python object instead of a dict
+      loaded_advisory: AdvisoryInfo = json.loads(
+          handle.read(), object_hook=lambda d: SimpleNamespace(**d))
+      if not hasattr(loaded_advisory,
+                     'modified') or not loaded_advisory.modified:
+        continue
 
-    print('Complete')
+      loaded_advisory.preexisting = True
```
oliverchang, line 321: nit: this adds a fair bit of complexity, and is not really necessary is it? we can remove it if so. 

I expect we'll be running this from scratch most of the time.

tools/debian/convert_debian.py
```diff
@@ -51,227 +56,307 @@
...
+    self.published = ''
+    self.details = ''
+    self.preexisting = False
+
+  def to_dict(self):
+    result = copy.copy(self.__dict__)
+    # Remove the preexisting key from the json output since
+    # it's only for internal use in the script
+
+    result.pop('preexisting')
```
oliverchang, line 152: See below: I think we can just do away with this completely.

tools/debian/first_package_finder.py
```diff
@@ -0,0 +1,120 @@
+"""Functions to help find the first package version for a specific release"""
```
oliverchang, line 1: As discussed let's add a main() to this and make this a standalone script that dumps the results in a JSON.

tools/debian/convert_debian.py
```diff
@@ -51,227 +56,307 @@
...
+      # Split by ',' here for the occasional case where there
+      # are two dates in the 'publish' field.
+      # Multiple dates are caused by major modification later on.
+      # This is accounted for with the modified timestamp with git
+      # below though, so we don't need to parse them here
+      advisory.published = (
+          datetime.datetime.strptime(report_date.split(',')[0],
+                                     '%Y-%m-%d').isoformat() + 'Z')
+
+    # git_relative_path = pathlib.Path(val_data).relative_to(webwml_repo)
```
oliverchang, line 275: nit: add a TODO here to fix/re-enable this.

oliverchang (commented):
nice! some initial comments.

tools/debian/first_package_finder.py
another-rex: I do like having the if and the else as the last section of a function, I feel like it makes it clearer that these are the two return paths.

tools/debian/first_package_finder.py
```diff
@@ -0,0 +1,134 @@
...
+import gzip
+
+import pandas as pd
+
+DEBIAN_RELEASE_VERSIONS_URL = 'https://debian.pages.debian.net/distro-info-data/debian.csv'
+
+
+def create_url(date: datetime, version: str = None):
+  """Create an url for snapshot.debian.org"""
+  base_url = 'https://snapshot.debian.org/archive/debian/{date}/dists/'
```
oliverchang, line 14: Put these in global constants above.

tools/debian/first_package_finder.py
```diff
@@ -0,0 +1,134 @@
...
+
+def parse_first_seen_dates(date: str) -> datetime:
+  """Parse first seen date in debian table to datetime"""
+  return datetime.strptime(date, '%Y-%m-%d %H:%M:%S')
+
+
+def fillout_first_seen(date: datetime, first_seen_dict: dict[str, datetime]):
+  """Fill out first seen version dict"""
+  with request.urlopen(create_url(date)) as result:
+    # Pandas will try to convert every table on the webpage to a dataframe.
```
oliverchang, line 54: This is neat but potentially brittle..

Could we just use the first created date as listed on https://debian.pages.debian.net/distro-info-data/debian.csv?

oliverchang (commented):
nice!!

tools/debian/first_package_finder.py
another-rex: Renamed to `get_debian_dists_url` and `get_debian_sources_url`

tools/debian/first_package_finder.py
another-rex: We can, the main issue is that we won't be getting the very first Source file, but probably the snapshot 24 - 36 hours later as I need to get the snapshot of the next day after the date listed there to reliably find it (though we can increase complexity and try a couple of different times around the listed date to make sure we get the first snapshot if this is very important).

Either case, I agree the current method is a bit brittle, so I'll put the current method in the alternative's we've considered part of the design doc.

another-rex: Just implemented it using debian.csv. Because debian snapshot api will always retrieve the previous snapshot for each date that is passed in, if there is a break in snapshots around when a release is made (for example when squeeze is released, there was over a week without snapshots), it won't find the first snapshot. 

This could potentially be fixed by looking forward a couple days until it does find a snapshot containing the release.

oliverchang: OK, let's go with this. retrying by walking a few days doesn't sound too bad, and is much more reliable than parsing the html table.

tools/debian/first_package_finder.py
```diff
@@ -0,0 +1,129 @@
...
+    # So it is safe to return 0, indicating the earliest version
+    # given by the snapshot API
+    return '0'
+
+
+def main():
+  dataframe = load_first_packages()
+  version_to_sources = dict(zip(dataframe['version'], dataframe['sources']))
+  with open('first_package_cache.json.gz', 'wb') as output_file:
+    result = gzip.compress(bytes(json.dumps(version_to_sources), 'utf-8'))
```
oliverchang, line 124: nit: use `json.dumps(...).encode('utf-8')` instead

tools/debian/first_package_finder.py
```diff
@@ -0,0 +1,129 @@
...
+
+  first_seen_dates = zip(
+      codename_to_version.index,
+      codename_to_version['created'].map(parse_created_dates_and_set_time))
+
+  first_seen_dict: dict[str, datetime] = dict(x for x in first_seen_dates if x[0] not in IGNORED_DEBIAN_VERSIONS)
+
+  for version, dates in first_seen_dict.items():
+    # retry for 10 days into the future
+    for i in range(10):
```
oliverchang, line 91: nit: put 10 in a global constant. i.e. `FIRST_SEEN_LOOKAHEAD_DAYS`

tools/debian/first_package_finder.py
```diff
@@ -0,0 +1,129 @@
...
+  """Parse created date in debian version csv to datetime plus one day"""
+  result = datetime.strptime(date, '%Y-%m-%d') + timedelta(days=1)
+  # Set minimum date to first debian snapshot
+  return max(result, datetime(2005, 3, 12))
+
+
+def load_sources(date: datetime, dist: str) -> dict[str, str]:
+  """Load the sources file and store in a dictionary of {name: version}"""
+  with request.urlopen(get_debian_sources_url(date, dist)) as res:
+    decompressed = gzip.decompress(res.read()).decode('utf-8', errors='ignore')
```
oliverchang, line 63: errors='ignore' seems dangerous. did you notice any issues without this specified?

tools/debian/first_package_finder.py
```diff
@@ -0,0 +1,129 @@
...
+
+def get_debian_dists_url(date: datetime):
+  """Create an url for snapshot.debian.org to get distribution"""
+  formatted_date = convert_datetime_to_str_datetime(date)
+  return DEBIAN_SNAPSHOT_URL.format(date=formatted_date)
+
+
+def get_debian_sources_url(date: datetime, version: str):
+  """Create an url for snapshot.debian.org"""
+
```
oliverchang, line 27: nit: no need for this newline.

tools/debian/first_package_finder.py
```diff
@@ -0,0 +1,129 @@
...
+def load_first_packages() -> pd.DataFrame:
+  """Loads the dataframe containing the first version of packages per distro"""
+
+  codename_to_version: pd.DataFrame = retrieve_codename_to_version()
+
+  first_seen_dates = zip(
+      codename_to_version.index,
+      codename_to_version['created'].map(parse_created_dates_and_set_time))
+
+  first_seen_dict: dict[str, datetime] = dict(x for x in first_seen_dates if x[0] not in IGNORED_DEBIAN_VERSIONS)
```
oliverchang, line 87: nit: line is too long. we should keep things under 80 columns.

I think you can just move the part after the `dict(` into a newline.

tools/debian/first_package_finder.py
```diff
@@ -0,0 +1,129 @@
...
+
+  for version, dates in first_seen_dict.items():
```
oliverchang, line 89: s/dates/date/ ? Should this be singular ?

tools/debian/first_package_finder.py
```diff
@@ -0,0 +1,129 @@
...
+        codename_to_version.loc[version].sources = load_sources(actual_date, version)
+        break
+      except urllib.error.HTTPError as http_error:
+        if http_error.code != 404:
+          raise
+        # Expect 404 errors for releases before snapshot exists
+
+      if actual_date > datetime.utcnow():
+        # No need to keep trying future dates
+        continue
```
oliverchang, line 103: can we break here instead?

tools/debian/first_package_finder.py
```diff
@@ -0,0 +1,129 @@
...
+    codename_to_version.loc['sid']['version'] = 'unstable'
+
+  return codename_to_version
+
+
+def parse_created_dates_and_set_time(date: str) -> datetime:
+  """Parse created date in debian version csv to datetime plus one day"""
+  result = datetime.strptime(date, '%Y-%m-%d') + timedelta(days=1)
+  # Set minimum date to first debian snapshot
+  return max(result, datetime(2005, 3, 12))
```
oliverchang, line 57: nit: put datetime(2005, 3, 12) in a global CONSTANT.

tools/debian/first_package_finder.py
```diff
@@ -0,0 +1,129 @@
...
+import pandas as pd
+
+DEBIAN_RELEASE_VERSIONS_URL = \
+  'https://debian.pages.debian.net/distro-info-data/debian.csv'
+DEBIAN_SNAPSHOT_URL = 'https://snapshot.debian.org/archive/debian/{date}/dists/'
+# `.gz` format always exist for all snapshots
+DEBIAN_SOURCES_URL_EXTENSION = '{version}/main/source/Sources.gz'
+
+# List of ignored versions, mostly too early to be in snapshots
+IGNORED_DEBIAN_VERSIONS = ['experimental', 'buzz', 'rex', 'bo', 'hamm', 'slink', 'potato']
```
oliverchang, line 17: nit: use frozenset here

oliverchang (approved):
LGTM with some final comments. can merge after addressing :)

tools/debian/first_package_finder.py
another-rex: image
Some maintainer names are not proper unicode in the earlier snapshots. They seem to have fixed it in later versions for example:
image

tools/debian/first_package_finder.py
another-rex: Yep, not sure why I put continue there haha

tools/debian/first_package_finder.py
oliverchang: ack!

oliverchang (approved):

Issue #25, opened 2022-01-27
Discussion about references field


kurtseifried:
The references field which is currently:

{
	"references": [ {
		"type": string,
		"url": string
	} ]
}

One major change I'd like to propose is that type be EITHER a string (singular) or an array (multiple entries), although ideally I'd prefer it just be an array with 1 or more entries.

Related to this I'd like to propose some additional values for type such as:

EXPLOIT (exploit code/methodology/etc.)
EXPLOITATION (reports of exploitation)
WORKAROUND
DISCUSSION (e.g. Twitter threads)

and obviously, some of these URLs can be multiple things, e.g. an ADVISORY could contain FIX and WORKAROUND and EXPLOIT hence the string/array request

I'd also like to propose adding a timestamp for when it was seen, this is extremely useful when creating timelines and I've noticed a worrying trend for web pages/forum posts/etc to not contain a published date or "1 month ago".

As you can see from https://github.com/cloudsecurityalliance/gsd-database/blob/main/2021/1002xxx/GSD-2021-1002352.json (search for "timestamp") it's incredibly helpful when reconstructing the Twitter activity timeline.

oliverchang:
Adding these link types SGTM, but allowing `type` to be an array would be a potentially more breaking change. Do you have any examples of existing GSD data that requires multiple types per link?  Alternatively, the same link could be embedded multiple times with different types?

timestamps are interesting, but the date an article/post was published can be very different from when an event actually happened, leading to misleading timelines right? I'm a little wary of including this complexity in the current schema using `references`. Perhaps a separate `timeline` field would be better suited for this, if there's enough interest for that?

PR #50, opened 2022-06-24
Add comments for references and evidence type

joshbuker:
Opening this up for discussion

### DIFF
--- a/docs/schema.md
+++ b/docs/schema.md
@@ -92,7 +92,8 @@ A JSON Schema for validation is also available
 	} ],
 	"references": [ {
 		"type": string,
-		"url": string
+		"url": string,
+		"comment": string
 	} ],
 	"credits": [ {
 		"name": string,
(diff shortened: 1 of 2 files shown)


oliverchang:
Thanks for opening this @joshbuker !

As discussed in person, our goal with this schema was to keep as many fields machine readable as possible, and keep the core fields as minimal as possible (i.e. focus on the purpose of enabling vulnerability scanners and triage).

The "type" field was intended to provide this kind of context for references, with a consistent mapping to a text description for what they mean. 

However, I recognize that different databases may want to track additional information for humans/triage purposes, e.g. @kurtseifried also suggested adding timestamps here as part of https://github.com/ossf/osv-schema/issues/25. Our mechanism for doing this was `database_specific`, but it wasn't very generalized. 

Perhaps an alternative here might be to extend the places where `database_specific` could apply. i.e. it can go in **any** part of the OSV schema, rather than the explicitly listed locations. i.e. this might look like:

```json
{
  "references": [
    {
      "type": "WEB",
      "url": "https://blah.com",
      "database_specific": {
        "timestamp": "....",
        "comment": "GSD specific comment"
      }
    }
  ]
}
```

We can also build out a more well specified way to define the `database_specific` specs from all the databases. e.g. linking to a JSON schema/spec page somewhere that describes all the fields that have been extended by that database. 

This mechanism doesn't exclude any of these fields from being added as a core field in the future -- if enough databases use similar fields it makes a strong case to include this as a core field. 

@rsc @chrisbloom7 thoughts?

kurtseifried:
I think if we're going to have a tag that then contains JSON and is scattered all over the place in order to allow new/different data we should look at how to make this less painful. Some thoughts:

Assume we keep the name database_specific for now. I suggest we usually add some standard metadata to it, e.g. 

data_format (e.g. OSV, GSD, CVE, CSAF, whatever)
data_version (e.g. 1.2.3, 5.0, etc.)

So people know what schema in turn to look at rather than having to write magic parsers/etc.

We can also use it to ascertain if stuff should be included in the schema, e.g. if we see thousands of timestamp tags then maybe we should make that part of the official schema, e.g. how HTTP headers work (if enough people do it, it's a de facto standard).

joshbuker:
Having thought this through more over the last week or two, I think we look at being very quick to iterate on adding new categories/reference types, and try to avoid adding a prose field for references until the actual demand from folks using it comes up.

PR #55, opened 2022-07-01
Replace removeprefix since it doesn't exist in 3.8

another-rex:
Dockerfile runs python 3.8, so can't use removeprefix, which only got added in 3.9


tools/debian/first_package_finder.py
```diff
@@ -73,11 +73,11 @@ def load_sources(date: datetime, dist: str) -> typing.Dict[str, str]:
     current_package = None
     for line in decompressed.splitlines():
       if line.startswith('Package: '):
-        current_package = line.removeprefix('Package: ')
+        current_package = line[len('Package: '):]
```
oliverchang, line 76: let's put these into constants.

PR #56, opened 2022-07-04
Debian DLAs and remove not affected

another-rex:
Support importing debian DLAs via the --lts flag, and also remove <not-affected> versions and advisories with empty affected fields.

### DIFF
--- a/tools/debian/convert_debian.py
+++ b/tools/debian/convert_debian.py
@@ -55,6 +55,8 @@ WML_REPORT_DATE_PATTERN = re.compile(
 CAPTURE_DSA_WITH_NO_EXT = re.compile(r'dsa-\d+')
 CAPTURE_DLA_WITH_NO_EXT = re.compile(r'dla-\d+')
 
+NOT_AFFECTED_VERSION = '<not-affected>'
+
 
 class AffectedInfo:
   """Debian version info."""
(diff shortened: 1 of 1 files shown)


tools/debian/convert_debian.py
```diff
@@ -136,13 +141,13 @@ def dumper(obj):
 
 
 def parse_security_tracker_file(advisories: Advisories,
-                                security_tracker_repo: str):
+                                security_tracker_repo: str, lts: bool):
   """Parses the security tracker files into the advisories object"""
 
   codename_to_version = create_codename_to_version()
-
+  security_tracker_path = SECURITY_TRACKER_DLA_PATH if lts else SECURITY_TRACKER_DSA_PATH
```
oliverchang, line 148: in general it's better to make functions like these more general and avoid hardcoding cases like these.

Let's have this function take in a string path instead and have the caller make the decision about which path to read from.

tools/debian/convert_debian.py
```diff
@@ -246,6 +256,9 @@ def parse_webwml_files(advisories: Advisories, webwml_repo: str):
 def write_output(output_dir: str, advisories: Advisories):
   """Writes the advisory dict into individual json files"""
   for dsa_id, advisory in advisories.items():
+    # Skip advisories that does not affect anything
+    if len(advisory.affected) == 0:
```
oliverchang, line 260: let's log this case

oliverchang (approved):
nice, thanks! LGTM with just some minor comments

oliverchang (approved):

PR #58, opened 2022-07-05
Add reference entry to outputted json

another-rex:
Add the url of the debian security advisory for each osv json entry

### DIFF
--- a/tools/debian/convert_debian.py
+++ b/tools/debian/convert_debian.py
@@ -30,7 +30,7 @@ WEBWML_SECURITY_PATH = os.path.join('english', 'security')
 WEBWML_LTS_SECURITY_PATH = os.path.join('english', 'lts', 'security')
 SECURITY_TRACKER_DSA_PATH = os.path.join('data', 'DSA', 'list')
 SECURITY_TRACKER_DLA_PATH = os.path.join('data', 'DLA', 'list')
-DEBIAN_BASE_PATH = "https://www.debian.org"
+DEBIAN_BASE_URL = "https://www.debian.org"
 
 LEADING_WHITESPACE = re.compile(r'^\s')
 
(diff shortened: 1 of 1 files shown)


tools/debian/convert_debian.py
```diff
@@ -30,6 +30,7 @@
 WEBWML_LTS_SECURITY_PATH = os.path.join('english', 'lts', 'security')
 SECURITY_TRACKER_DSA_PATH = os.path.join('data', 'DSA', 'list')
 SECURITY_TRACKER_DLA_PATH = os.path.join('data', 'DLA', 'list')
+DEBIAN_BASE_PATH = "https://www.debian.org"
```
oliverchang, line 33: nit: s/PATH/URL/

tools/debian/convert_debian.py
```diff
@@ -237,6 +251,15 @@ def parse_webwml_files(advisories: Advisories, webwml_repo: str,
           datetime.datetime.strptime(report_date.split(',')[0],
                                      '%Y-%m-%d').isoformat() + 'Z')
 
+    advisory_url = '/'.join([DEBIAN_BASE_PATH,
+                             # Remove webwml/english/ from the path
+                             # And also remove the .wml extension
+                             *val_wml.split('/')[2:]])[:-len('.wml')]
```
oliverchang, line 257: nit: this is a little hard to read, and doesn't work when the "webwml" checkout directory lives in a different directory right? (as opposed to be relative to the script's CWD). 

let's do something like this: 

```python
advisory_url_path = os.path.relpath(wml_path, os.path.join(webwml_repo_path, 'english'))
# Remove .whl extension.
advisory_url_path = os.path.splitext(advisory_url_path)[0]
advisory_url = DEBIAN_BASE_URL + '/' + advisory_url_path
```

Also, I missed this in the earlier review, but while we're here could we rename `val_wml` to `wml_path` and `val_data` to `data_path` to make it clearer what these variables are referring to? We can rename `webwml_path` to `webwml_repo_path` to disambiguate them better.

oliverchang (approved):

PR #59, opened 2022-07-07
add modified date entry to convert_debian output from git history

another-rex:
The git repository for webwml must be the full repository, not `--depths=n` for modified date to be not null


oliverchang (approved):
very nice work!

PR #60, opened 2022-07-08
Fix modified date format (removes the +00:00 which fails worker tests)

another-rex:


tools/debian/convert_debian.py
```diff
@@ -283,7 +284,8 @@ def parse_webwml_files(advisories: Advisories, webwml_repo_path: str,
 
     if line.startswith(GIT_DATE_PREFIX):
       current_date = datetime.datetime.fromisoformat(
-          line[len(GIT_DATE_PREFIX):]).astimezone(datetime.timezone.utc).isoformat() + 'Z'
+          line[len(GIT_DATE_PREFIX):]).astimezone(
```
oliverchang, line 287: Can you just add a comment here saying the spec requires a "Z" offset,

oliverchang (approved):

PR #63, opened 2022-07-11
Also check for modification of .wml files (the description)

another-rex:
We were only checking for .data file modifications (mainly because of performance cost), now with the new method also checking for .wml file modifications and getting the latest modification have minimal performance cost.

### DIFF
--- a/tools/debian/convert_debian.py
+++ b/tools/debian/convert_debian.py
@@ -13,11 +13,12 @@
 # limitations under the License.
 """Debian to OSV converter."""
 import argparse
+import collections
 import io
 import json
 import os
 import re
-import datetime
+from datetime import datetime, timezone
 import subprocess
 import typing
 from urllib import request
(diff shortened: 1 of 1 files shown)


tools/debian/convert_debian.py
```diff
@@ -265,9 +266,13 @@ def parse_webwml_files(advisories: Advisories, webwml_repo_path: str,
...
 
-    git_relative_path = os.path.relpath(data_path, webwml_repo_path)
-    git_relative_data_paths[git_relative_path] = dsa_id
+    git_relative_path_wml = os.path.relpath(wml_path, webwml_repo_path)
+    git_relative_path_data = os.path.relpath(data_path, webwml_repo_path)
+    git_relative_paths[git_relative_path_wml] = dsa_id
+    git_relative_paths[git_relative_path_data] = dsa_id
 
+  modified_date_dict = collections.defaultdict(
+    lambda: datetime.fromtimestamp(0, timezone.utc))
```
oliverchang, line 275: Is this indented correctly? Typically these line continuations would need double (i.e. 4 spaces) here.

oliverchang (approved):
nice!

tools/debian/convert_debian.py
another-rex: Ah, nice catch, ran pyformat now. I think I should to put pyformat on a pre commit hook to have it run automatically

PR #70, opened 2022-07-13
Add support for DTSA

another-rex:
Adds support for importing DTSA files.

The dateutil dependency is already included in the lock file from another dependency, so no lock file changes

### DIFF
--- a/tools/debian/convert_debian.py
+++ b/tools/debian/convert_debian.py
@@ -22,7 +22,9 @@ from datetime import datetime, timezone
 import subprocess
 import typing
 from urllib import request
+from enum import Enum
 
+import dateutil.parser
 import markdownify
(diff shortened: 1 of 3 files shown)


tools/debian/convert_debian.py
```diff
@@ -360,13 +382,16 @@ def main():
   parser.add_argument(
       '-o', '--output-dir', help='Output directory', required=True)
   parser.add_argument(
-      '--lts', help='Convert long term support advisories', action='store_true')
+      '--adv_type',
```
oliverchang, line 385: Can you also update the README with some documentation?

oliverchang (approved):
thanks!!

oliverchang (approved):

PR #72, opened 2022-07-19
First package finder that exports to individual files instead packagi…

another-rex:
First package finder that exports to individual files instead packaging everything into a single file.

Also fixes formatting that was broken from a previous commit, and copies the yapf style file into the git repository

### DIFF
--- a/tools/debian/first_package_finder.py
+++ b/tools/debian/first_package_finder.py
@@ -1,4 +1,6 @@
 """Functions to help find the first package version for a specific release"""
+import argparse
+import os
 import typing
 import urllib.error
 from datetime import datetime, timedelta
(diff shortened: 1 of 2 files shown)


tools/debian/first_package_finder.py
```diff
@@ -30,112 +32,128 @@
...
+  version_to_sources = dict(zip(dataframe['version'], dataframe['sources']))
+
+  for version, sources in version_to_sources.items():
+    versions_path = os.path.join(args.output_dir, version)
+    os.makedirs(versions_path, exist_ok=True)
+    if not sources:
+      continue
+
+    for package, versions in sources.items():
+      with open(os.path.join(versions_path, package + '.txt'),
```
oliverchang, line 153: nit: no need for the ".txt" extension.

oliverchang (approved):
nice!

PR #74, opened 2022-07-26
Add GitHub Actions and Pub to the ecosystems list

rthorpeii:
This adds [GitHub Actions](https://docs.github.com/en/actions) and [Pub](https://pub.dev/) to the documented list of defined ecosystems in order to support GitHub's security advisories.

There are no validations on the ecosystems defined in the schema docs, so the validation JSON has not been updated.

cc: @katblag

### DIFF
--- a/docs/schema.md
+++ b/docs/schema.md
@@ -382,7 +382,7 @@ The defined ecosystems are:
 | `Debian` | The Debian package ecosystem; the `name` is the name of the source package. The ecosystem string might optionally have a `:<RELEASE>` suffix to scope the package to a particular Debian release. `<RELEASE>` is a numeric version specified in the [Debian distro-info-data](https://debian.pages.debian.net/distro-info-data/debian.csv). For example, the ecosystem string "Debian:7" refers to the Debian 7 (wheezy) release.  |
 | `Hex` | The package manager for the Erlang ecosystem; the `name` is a Hex package name.  |
 | `Android`  | The Android ecosystem; the `name` field is the Android component name that the patch applies to, as shown in the [Android Security Bulletins](https://source.android.com/security/bulletin) such as `Framework`, `Media Framework` and `Kernel Component`. The exhaustive list of components can be found at the [Appendix](#android-ecosystem-components). |
-| `GitHub Actions` | The GitHub Actions ecosystem; the `name` field is an action name. |
+| `GitHub Actions` | The GitHub Actions ecosystem; the `name` field is the action's repository name with owner e.g. `{owner}/{repo}`. |
 | `Pub` | The package manager for the Dart ecosystem; the `name` field is a Dart package name. |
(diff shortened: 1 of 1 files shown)


docs/schema.md
```diff
@@ -382,6 +382,8 @@ The defined ecosystems are:
 | `Debian` | The Debian package ecosystem; the `name` is the name of the source package. The ecosystem string might optionally have a `:<RELEASE>` suffix to scope the package to a particular Debian release. `<RELEASE>` is a numeric version specified in the [Debian distro-info-data](https://debian.pages.debian.net/distro-info-data/debian.csv). For example, the ecosystem string "Debian:7" refers to the Debian 7 (wheezy) release.  |
 | `Hex` | The package manager for the Erlang ecosystem; the `name` is a Hex package name.  |
 | `Android`  | The Android ecosystem; the `name` field is the Android component name that the patch applies to, as shown in the [Android Security Bulletins](https://source.android.com/security/bulletin) such as `Framework`, `Media Framework` and `Kernel Component`. The exhaustive list of components can be found at the [Appendix](#android-ecosystem-components). |
+| `GitHub Actions` | The GitHub Actions ecosystem; the `name` field is an action name. |
```
oliverchang, line 385: Are there any other rules for how this name should be formated? 

Is it a string such as "google/clusterfuzzlite/actions/build_fuzzers" ? (No @ part?). It may be helpful to clarify this.

docs/schema.md
```diff
@@ -382,6 +382,8 @@ The defined ecosystems are:
...
+| `Pub` | The package manager for the Dart ecosystem; the `name` field is a Dart package name. |
```
oliverchang, line 386: @jonasfj Does this seem good to you? Are there any normalization rules etc for Pub packages?

oliverchang (approved):
Thank you for adding these! (And very sorry for the slow review -- I've been out travelling).

This LGTM with just some minor questions.

docs/schema.md
jonasfj: This looks solid to me.

There is some corner case around custom pub repositories. But that's probably solved with a `purl` that has a `?repository_url=...` qualifier.

jonasfj (approved):

docs/schema.md
rthorpeii: Thanks for the feedback! We've added some more info for clarity.

oliverchang: Thanks a lot! Sorry one more question -- would it make sense for this to be `owner/repo/path` where `path` is optional? Or is the granularity going to be at the repo level?

rthorpeii: Good question, and one that we had ourselves. We synced with the Actions team on this and determined it was best to leave it as `owner/repo` to better align with the direction that Actions are headed, and Actions that currently follow `owner/repo/path` can have path captured at the repo level

oliverchang: ack, thanks!

PR #87, opened 2022-09-09
Added CVSS_V2 and fixed whitespace

kurtseifried:

### DIFF
--- a/docs/schema.md
+++ b/docs/schema.md
@@ -226,7 +226,7 @@ same database, the duplicate entry could be written using only the `id`,
 ```
 
 The `related` field gives a list of IDs of closely related vulnerabilities, such
-as the same problem in alternate ecosystems. 
+as the same problem in alternate ecosystems.
 
 ## summary, details fields
 
@@ -242,7 +242,7 @@ vulnerability. It is recommended that this field be kept short, on the order of
 no more than 120 characters.
 
 The `details` field gives additional English textual details about the
-vulnerability. 
+vulnerability.
 
 The `summary` field is plain text.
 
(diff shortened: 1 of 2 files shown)


docs/schema.md
```diff
@@ -276,6 +276,7 @@ describes the quantitative method used to calculate the associated `score`.
 
 | Severity Type | Score Description |
 | --------- | ----------- |
+| `CVSS_V2` | A CVSS vector string representing the unique characteristics and severity of the vulnerability using a version of the [Common Vulnerability Scoring System notation](https://www.first.org/cvss/v2/) that is >= 2.0 and < 3.0 (e.g.`"AV:L/AC:M/Au:N/C:N/I:P/A:C"`).|
```
oliverchang, line 279: Is CVSS 2.0 the only available version? I ask because the CVSS string itself does not seem to have any version information as part of it, making it impossible to distinguish where it likes in between 2.0 and 3.0. 

If 2.0 is the only version -- let's just call that out as the only version?

oliverchang:
Adding @chrisbloom7 for review too :) Since you added the CVSS_V3 support.

docs/schema.md
kurtseifried: Correct, That's one reason they added the version string in 3.x, in case it ever got upgraded. There is a 1, 2, 3.0 and 3.1, and a 4 planned...

kurtseifried: For reference: https://www.first.org/cvss/v2/history also the sidebar shows it, 1, 2, 3.0, 3.1.

kurtseifried: Technically just "2". They lack precission? =)

chrisbloom7 (approved):
👍🏻 ❤️

docs/schema.md
oliverchang: Hmm, https://www.first.org/cvss/v2/guide says "2.0" in the title, but I don't have strong opinions about either. Could you please update the text here? Then we can merge.

kurtseifried: I've only seen it referred to in urls/titles as v2 but google seems to find more "cvss 2.0" than "cvss 2"

kurtseifried: so what I'm saying is leave it as 2.0 (who knows, maybe one day there is a 2.1?)

oliverchang: OK -- can we update the text to say: 

```
that is == 2.0
```

instead of

```
that is >= 2.0 and < 3.0 
```

kurtseifried: ok done

oliverchang (approved):
thank you!

PR #88, opened 2022-10-05
Add alpine as a ecosystem entry in the osv schema

another-rex:
Add alpine as a ecosystem entry in the osv schema

### DIFF
--- a/docs/schema.md
+++ b/docs/schema.md
@@ -381,7 +381,7 @@ The defined ecosystems are:
 | `NuGet` | The NuGet package ecosystem. The `name` field is a NuGet package name.  |
 | `Linux` | The Linux kernel. The only supported `name` is `Kernel`. |
 | `Debian` | The Debian package ecosystem; the `name` is the name of the source package. The ecosystem string might optionally have a `:<RELEASE>` suffix to scope the package to a particular Debian release. `<RELEASE>` is a numeric version specified in the [Debian distro-info-data](https://debian.pages.debian.net/distro-info-data/debian.csv). For example, the ecosystem string "Debian:7" refers to the Debian 7 (wheezy) release.  |
-| `Alpine` | The Alpine package ecosystem; the `name` is the name of the source package. The ecosystem string might optionally have a `:<RELEASE>` suffix to scope the package to a particular Alpine release branch. E.g. `v3.16` |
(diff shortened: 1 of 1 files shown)


docs/schema.md
```diff
@@ -381,6 +381,7 @@ The defined ecosystems are:
 | `NuGet` | The NuGet package ecosystem. The `name` field is a NuGet package name.  |
 | `Linux` | The Linux kernel. The only supported `name` is `Kernel`. |
 | `Debian` | The Debian package ecosystem; the `name` is the name of the source package. The ecosystem string might optionally have a `:<RELEASE>` suffix to scope the package to a particular Debian release. `<RELEASE>` is a numeric version specified in the [Debian distro-info-data](https://debian.pages.debian.net/distro-info-data/debian.csv). For example, the ecosystem string "Debian:7" refers to the Debian 7 (wheezy) release.  |
+| `Alpine` | The Alpine package ecosystem; the `name` is the name of the source package. The ecosystem string might optionally have a `:<RELEASE>` suffix to scope the package to a particular Alpine release branch. E.g. `v3.16` |
```
oliverchang, line 384: Is there a definitely list/rules around how to refer to an Alpine release? i.e. the string has to match a branch name, or some rules around the 'v' prefix?

oliverchang (commented):
sorry for the delay on this one! lost track of this after I got back from vacation.

docs/schema.md
another-rex: I can't find any. The git branches in aports do not have the v prefix, but the release branches here: https://alpinelinux.org/releases/ do

oliverchang: In that case, can we be clear about the "v" prefix in the spec here?

kaniini: The `v` should be explicit I guess.  The repositories are named `vMAJOR.MINOR` on the mirrors.

oliverchang: Thanks @kaniini ! @another-rex can you please make this 'v' part more explicit? 

Are there any other subtle rules to how to uniquely identify an Alpine package? Also, given an Alpine `(package name, version number)`, is that necessarily unique across all Alpine branches? Or is it possible that the same `(package name, version number)` can be different depending on the release branch?

kaniini: It is possible that (package name, version number) is different on different release branches.

oliverchang (approved):

docs/schema.md
oliverchang: Thanks @kaniini ! @another-rex can we make this fact clearer, and change the "might optionally have a" to a "must have a" ?

oliverchang: (I just did this).

PR #89, opened 2022-10-06
Add Evidence and Discussion reference types

### LINKED ISSUE #78: Additional options for references field to support reproducers/exploit code

Ok meta comments:

joshbuker:
Also normalized the ordering for the types to be alphabetical order.

Closes #78

### DIFF
--- a/docs/schema.md
+++ b/docs/schema.md
@@ -690,14 +690,17 @@ The known reference `type` values are:
 
 - `ADVISORY`: A published security advisory for the vulnerability.
 - `ARTICLE`: An article or blog post describing the vulnerability.
-- `REPORT`: A report, typically on a bug or issue tracker, of the vulnerability.
+- `DETECTION`: An article or code that provides detection of the vulnerability.
+- `DISCUSSION`: A discussion of the vulnerability and related issues, e.g.
+	Twitter, Reddit.
+- `EVIDENCE`: A demonstration of the validity of a vulnerability claim, e.g.
+	`app.any.run` replaying the exploitation of the vulnerability.
 - `FIX`: A source code browser link to the fix (e.g., a GitHub commit) Note that
   the `fix` type is meant for viewing by people using web browsers.  Programs
   interested in analyzing the exact commit range would do better to use the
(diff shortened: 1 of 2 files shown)


docs/schema.md
```diff
@@ -690,14 +690,17 @@ The known reference `type` values are:
 
 - `ADVISORY`: A published security advisory for the vulnerability.
 - `ARTICLE`: An article or blog post describing the vulnerability.
-- `REPORT`: A report, typically on a bug or issue tracker, of the vulnerability.
+- `DETECTION`: An article or code that provides detection of the vulnerability.
```
oliverchang, line 693: What does "article or code that provides detection" mean exactly? 

It seems a bit ambiguous to me -- does it mean:

- "Code" that reproduces the issue (in which case `EVIDENCE` seems like a better fit?), or 
- An article that talks about how this vuln was discovered?

oliverchang (commented):
Thanks! Apologies again for the delay -- I was out on vacation for a few weeks.

docs/schema.md
chrisbloom7: I agree. It sounds like this could already be represented by either `ARTICLE` or `EVIDENCE`.

joshbuker: @kurtseifried Do you want to expand on this?

kurtseifried: so for example YARA rules, semgrep, etc. e.g. https://gsd.id/GSD-2021-44228 

est detection:

grep: https://gist.github.com/Neo23x0/e4c8b03ff8cdf1fa63b7d15db6e3860b

jarhashes: https://github.com/mubix/CVE-2021-44228-Log4Shell-Hashes

semgrep: https://github.com/returntocorp/semgrep-rules/pull/1650/commits/ecfc32623eec718d61ec83b9196574f333191008/

yara: https://github.com/timb-machine/log4j/

burpsuite: https://github.com/silentsignal/burp-log4shell

Nmap NSE: https://github.com/Diverto/nse-log4shell

joshbuker:
https://github.com/cloudsecurityalliance/gsd-tools/pull/197

PR #93, opened 2022-11-01
Loosen schema requirement for versions array

### LINKED ISSUE #79: versions required if git commit present, but versions can be blank []
the versions array is mandatory if you e.g. put a git commit in for affected, but the versions array can be blank:

   "allOf": [
          {
            "if": {
              "properties": {
                "ranges": {
                  "contains": {
                    "properties": {
                      "type": {
                        "enum": [
                          "SEMVER",
                          "ECOSYSTEM"
                        ]
                      }
                    }
                  }
                }
              }
            },
            "then": {},
            "else": {
              "required": [
                "versions"

I would ask it either be removed or if it's intended to be there, then some content should be checked for (which breaks things, some things have git commit's but no releases/etc.) so I would suggest we remove this versions requirement.

another-rex:
Loosen schema requirement for versions array, since it should be valid OSV to have just git commit ranges.

Fixes #79

Signed-off-by: Rex P <106129829+another-rex@users.noreply.github.com>

### DIFF
--- a/validation/schema.json
+++ b/validation/schema.json
@@ -230,7 +230,8 @@
                       "type": {
                         "enum": [
                           "SEMVER",
-                          "ECOSYSTEM"
+                          "ECOSYSTEM",
+                          "GIT"
                         ]
                       }
                     }
--- a/validation/schema.json
+++ b/validation/schema.json
@@ -219,34 +219,7 @@
           "database_specific": {
             "type": "object"
           }
-        },
-        "allOf": [
-          {
-            "if": {
-              "properties": {
-                "ranges": {
-                  "contains": {
-                    "properties": {
-                      "type": {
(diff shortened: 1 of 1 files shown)


validation/schema.json
```diff
@@ -230,7 +230,8 @@
                       "type": {
                         "enum": [
                           "SEMVER",
-                          "ECOSYSTEM"
+                          "ECOSYSTEM",
```
oliverchang, line 233: we can just get rid of the entire allOf, since SEMVER, ECOSYSTEM, and GIT encompass every single type.

another-rex: Good point, removed

oliverchang (approved):

oliverchang (approved):

PR #95, opened 2022-11-08
Allow null instead of an empty array.

another-rex:
Some of the existing entries use null instead of being an empty array or removing the entry entirely.

### DIFF
--- a/validation/schema.json
+++ b/validation/schema.json
@@ -276,10 +276,7 @@
     }
   },
   "required": [
-    "id"
-  ],
-  "anyOf": [
-    { "required": ["modified"] },
-    { "required": ["withdrawn"] }
+    "id",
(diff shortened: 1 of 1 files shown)


oliverchang (approved):
LGTM. Do other non-array fields also allow `null` values with the current schema?

another-rex:
They do not, should we allow null values for them as well?

Running through all the currently imported cves, none of them have null in other fields, but maybe we should be consistent in allowing nulls.

oliverchang:
> They do not, should we allow null values for them as well?
> 
> Running through all the currently imported cves, none of them have null in other fields, but maybe we should be consistent in allowing nulls.

We can start with this for now.

oliverchang (approved):

PR #101, opened 2022-12-19
- add conan.io ecosystem

SSE4:
add [conan.io](https://conan.io) ecosystem as an ecosystem entry in the ovs schema.

### DIFF
--- a/docs/schema.md
+++ b/docs/schema.md
@@ -388,7 +388,7 @@ The defined ecosystems are:
 | `Android`  | The Android ecosystem; the `name` field is the Android component name that the patch applies to, as shown in the [Android Security Bulletins](https://source.android.com/security/bulletin) such as `Framework`, `Media Framework` and `Kernel Component`. The exhaustive list of components can be found at the [Appendix](#android-ecosystem-components). |
 | `GitHub Actions` | The GitHub Actions ecosystem; the `name` field is the action's repository name with owner e.g. `{owner}/{repo}`. |
 | `Pub` | The package manager for the Dart ecosystem; the `name` field is a Dart package name. |
-| `conan.io` | The conan.io ecosystem for C and C++; the `name` field is a conan package name.  |
+| `conan` | The conan ecosystem for C and C++; the `name` field is a conan package name.  |
 | Your ecosystem here. | [Send us a PR](https://github.com/ossf/osv-schema/compare). |
 
 It is permitted for a database name (the DB prefix in the `id` field) and an
--- a/docs/schema.md
+++ b/docs/schema.md
@@ -388,7 +388,7 @@ The defined ecosystems are:
 | `Android`  | The Android ecosystem; the `name` field is the Android component name that the patch applies to, as shown in the [Android Security Bulletins](https://source.android.com/security/bulletin) such as `Framework`, `Media Framework` and `Kernel Component`. The exhaustive list of components can be found at the [Appendix](#android-ecosystem-components). |
 | `GitHub Actions` | The GitHub Actions ecosystem; the `name` field is the action's repository name with owner e.g. `{owner}/{repo}`. |
 | `Pub` | The package manager for the Dart ecosystem; the `name` field is a Dart package name. |
-| `conan` | The conan ecosystem for C and C++; the `name` field is a conan package name.  |
+| `ConanCenter` | The ConanCenter ecosystem for C and C++; the `name` field is a Conan package name.  |
 | Your ecosystem here. | [Send us a PR](https://github.com/ossf/osv-schema/compare). |
 
 It is permitted for a database name (the DB prefix in the `id` field) and an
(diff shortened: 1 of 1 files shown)


docs/schema.md
```diff
@@ -388,6 +388,7 @@ The defined ecosystems are:
 | `Android`  | The Android ecosystem; the `name` field is the Android component name that the patch applies to, as shown in the [Android Security Bulletins](https://source.android.com/security/bulletin) such as `Framework`, `Media Framework` and `Kernel Component`. The exhaustive list of components can be found at the [Appendix](#android-ecosystem-components). |
 | `GitHub Actions` | The GitHub Actions ecosystem; the `name` field is the action's repository name with owner e.g. `{owner}/{repo}`. |
 | `Pub` | The package manager for the Dart ecosystem; the `name` field is a Dart package name. |
+| `conan.io` | The conan.io ecosystem for C and C++; the `name` field is a conan package name.  |
```
oliverchang, line 391: We typically stylize ecosystems the way it's typically referred to in canonical documentation. 

E.g. "crates.io" per https://www.rust-lang.org/governance/teams/crates-io. PyPI per https://pypi.org.

Looking at https://conan.io, should this be "Conan", or perhaps "ConanCenter" instead?

SSE4: conan documentation doesn't mention ecosystem at all.
google has more results for `conan ecosystem` than `conan.io ecosystem` or `conancenter ecosystem`. so I suppose `conan ecosystem` is canonical one.
`conan.io` is just FQDN. `ConanCenter` is a place where packages built and stored. it's a part of the ecosystem for sure, but it won't work without other things (like conan client).
@prince-chrismc lemme know if you agree.

SSE4, line 391:
```suggestion
| `conan` | The conan ecosystem for C and C++; the `name` field is a conan package name.  |
```

oliverchang: Thanks! We generally respect capitalization, so if we go with "Conan" the C should be capitalized. 

Can you explain a bit more about why `ConanCenter` doesn't work? The intention of these ecosystems is to unambiguously identify a package that exists in some open repository. If there are multiple repositories, then `Conan` on its own could be ambiguous because it's not clear to which repository a package belongs.

SSE4: conan is distributed by design and provides multiple repositories (remotes). e.g. everyone can easily run their own conan server. ConanCenter is just one of them.
I think it makes sense if we want to target specifically ConanCenter packages, to avoid ambiguity.
so I am changing to capiital `ConanCenter` then.

docs/schema.md
```diff
@@ -388,6 +388,7 @@ The defined ecosystems are:
 | `Android`  | The Android ecosystem; the `name` field is the Android component name that the patch applies to, as shown in the [Android Security Bulletins](https://source.android.com/security/bulletin) such as `Framework`, `Media Framework` and `Kernel Component`. The exhaustive list of components can be found at the [Appendix](#android-ecosystem-components). |
 | `GitHub Actions` | The GitHub Actions ecosystem; the `name` field is the action's repository name with owner e.g. `{owner}/{repo}`. |
 | `Pub` | The package manager for the Dart ecosystem; the `name` field is a Dart package name. |
+| `conan` | The conan ecosystem for C and C++; the `name` field is a conan package name.  |
```
SSE4, line 391:
```suggestion
| `ConanCenter` | The ConanCenter ecosystem for C and C++; the `name` field is a Conan package name.  |
```

docs/schema.md
oliverchang: Thanks! For the purposes of OSV, referring specifically to the main Conan repo for open source packages makes the most sense.

Do we want to wait for @prince-chrismc's input as well?

SSE4: no, I think we all have an agreement already

SSE4:
@oliverchang can we move this one forward? lemme know if you need anything else from me

oliverchang (approved):
Sorry for the delay! Approving.

andrewpollock:
@SSE4 is there an advisory source OSV.dev should be importing from?

SSE4:
> @SSE4 is there an advisory source OSV.dev should be importing from?

I am sorry, but I do not understand your question completely. importing what exactly and for what purpose? are we talking about python [import](https://docs.python.org/3/reference/simple_stmts.html#import) statement here, or something different? I have no idea what is "advisory source" in this context (I assume, you mean package source code repository?). I would be grateful if you add a little bit more context/background to your question.

/cc @memsharded & @jcar87 as conan devs (I personally no longer work for conan.io).

andrewpollock:
My apologies for the confusion, I have assumed prior knowledge. I am referring to importing advisories published in the OSV format, into https://OSV.dev so that they are searchable via the web interface, and discoverable via the API.

See also https://google.github.io/osv.dev/data/ and https://google.github.io/osv.dev/architecture/

memsharded:
Hi @andrewpollock 

I am afraid we don't have such a source for vulnerabilities of ConanCenter packages. This is something that would be interesting, but we don't have it yet.

andrewpollock:
I'm curious what the intent was behind getting the ecosystem added to the OSV schema?

SSE4:
> I'm curious what the intent was behind getting the ecosystem added to the OSV schema?

I was told to register conan ecosystem first here in order to proceed with this PR: https://github.com/google/osv-scanner/pull/59

Issue #40, opened 2022-03-28
Severity does not support multiple differing severities for one ID


joshbressers:
I just noticed if I have two different severities for the same identifier, I can't note this in a sensible manner.

For example

Let's say ID-1 affects two Linux distributions differently. On one distro it's critical, on the other it's important.

Right now the severity field is inflexible
https://ossf.github.io/osv-schema/#severity-field

It's an array, but the spec only support CVSSv3, and there is no way to namespace the findings into package or ecosystem. In the affected fields, I will specify an ecosystem to account for this
```
{"ecosystem": "distro1", "name": "package"}

and

{"ecosystem": "distro2", "name": "package"}
```

I've also seen instances where different versions have different severities to make this even more difficult. This is less common, so trying to put severity in the affected field would create a lot of duplication for what is not a common occurrence.

oliverchang:
Thanks for bringing this up! Not sure if there's a sensible way to encode the per-version severities, but perhaps we could add a per package "severity" field that overrides the top-level one. Is this something that you see often in practice?

joshbressers:
Putting severity as an optional field in "package" would work I think, yeah

kurtseifried:
I think there are two aspects here:

Nested data
Data viewpoints within a specific instance

So for example you have Red Hat Enterprise Linux 8, and 9, one is affected, one is not. Even within a version, you might have a default config that is not affected, and a common case config that is affected, thus very different severities. So I think nesting data e.g.:

Vendor
-Product
--Version 8
---Severity 8.2
--Version 9
---Config 1
----Severity 1.2
---Config2
----Severity 6.9

And then even within a specifically affected product/version/etc you may have different viewpoints on the severity, e.g. an information leak of just the user's first name may be viewed by many as Confidentiality: Moderate but by a regulated entity (e.g. due to privacy legislation) as Confidentiality: High. I think the way to express this is again with some version of nesting/namespacing within the data, e.g.:

Vendor
-Product
--Version 9
---Severity 1.2
---Viewpoint from X
----Severity 6.9

The logic generally being you traverse down the JSON and set values, and if you find a nested value it overides the already seen one.

oliverchang:
At some point, I think it makes more sense to split out separate vulnerability entries rather than try to encode all of this complexity in a single one. A Red Hat specific entry would have its own ID with its own opinions on severity. 

Re configurations, I think we need to all agree on a machine readable way to specify this. 

[...]

PR #106, opened 2023-01-05
Add optional severity field override for affected packages. 

oswalpalash:
Based on https://github.com/ossf/osv-schema/issues/40#issuecomment-1371719724

### DIFF
--- a/docs/schema.md
+++ b/docs/schema.md
@@ -73,12 +73,12 @@ A JSON Schema for validation is also available
 		"package": {
 			"ecosystem": string,
 			"name": string,
-			"purl": string,
-			"severity": [ {
-				"type": string,
-				"score": string
-			} ]
+			"purl": string
 		},
+		"severity": [ {
+			"type": string,
+			"score": string
+		} ],
 		"ranges": [ {
 			"type": string,
 			"repo": string,
@@ -299,12 +299,12 @@ on the selected `severity[].type`, as described above.
 		"package": {
 			"ecosystem": string,
 			"name": string,
-			"purl": string,
-			"severity": [ {
-				"type": string,
-				"score": string
-			} ]
+			"purl": string
 		},
+		"severity": [ {
+			"type": string,
+			"score": string
+		} ],
 		"ranges": [ {
 			"type": string,
 			"repo": string,
@@ -331,6 +331,12 @@ versions. In rare cases, for example if the `ecosystem_specific` encodes
 platform information that doesn't apply equally to all listed versions and
 ranges, a separate entry with the same `package` in the `affected` array may be
 needed.
+
+The `severity` field is an optional element [defined here](#severity-field). 
+This `severity` field applies to a specific package, in cases where affected 
+packages have differing severities for the same vulnerability. If any package 
+level `severity` fields are set, the top level [`severity`](#severity-field) 
+must not be set. 
 
 The `versions` field can enumerate a specific set of affected versions, and the
 `ranges` field can list ranges of affected versions, under a given defined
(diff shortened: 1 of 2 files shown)


captn3m0:
Looks good.

oliverchang (commented):
Thanks! 

I wonder if we need to call out any semantics here. i.e. What does it mean for there to be both a top level severity and per-package severity?

Does it make sense to have both? Or should we disallow specifying both? 

@joshbressers @kurtseifried @chrisbloom7 @rsc

docs/schema.md
```diff
@@ -356,6 +364,11 @@ The `purl` field is a string following the
  [Package URL specification](https://github.com/package-url/purl-spec) that
 identifies the package. This field is optional but recommended.
 
+The `severity` field is an optional JSON array that allows generating systems to 
+describe the severity of a vulnerability using one or more quantitative scoring 
+methods. Each `severity` item is a JSON object specifying a `type` and `score`
```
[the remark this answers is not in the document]
oswalpalash: @oliverchang I've made the changes.

chrisbloom7 (approved):

docs/schema.md
```diff
@@ -73,7 +73,11 @@ A JSON Schema for validation is also available
 		"package": {
 			"ecosystem": string,
 			"name": string,
-			"purl": string
+			"purl": string,
+			"severity": [ {
```
oliverchang, line 77: We should pull this one level up out of the `"package"` field into a field under "affected". The "package" field is meant for identifying the package. Any other properties go alongside it, not inside.

docs/schema.md
```diff
@@ -356,6 +364,12 @@ The `purl` field is a string following the
  [Package URL specification](https://github.com/package-url/purl-spec) that
 identifies the package. This field is optional but recommended.
 
+The `severity` field is an optional element [defined here](#severity-field). 
+Per package severity override is optional for different packages. It is only
```
oliverchang, line 368: Thinking more I'd suggest enforcing that either the top level severity or the per-package severites are set, but not both. 

It doesn't make much sense to me, and it enforces a bit more consistency. We should also validate this with our JSON schema, but that can come later. 

Maybe we can replace everything from the second paragraph onwards with:

```
This `severity` field applies to a specific package, in cases where affected packages have
differing severities for the same vulnerability. If any package level `severity` fields are set,
the top level `severity` (add hyperlink) must not be set. 
```

oliverchang (commented):
thanks again for this PR. a few more comments.

docs/schema.md
oswalpalash: Done! Ready for another review

docs/schema.md
```diff
@@ -324,6 +332,12 @@ platform information that doesn't apply equally to all listed versions and
 ranges, a separate entry with the same `package` in the `affected` array may be
 needed.
 
+The `severity` field is an optional element [defined here](#severity-field).
```
oliverchang, line 335: Sorry one more comment: Can you just move this text down to under the heading on line 411? That makes this text a bit more easily discoverable since it can be found from the ToC. 

The other text here around the relevant fields (e.g. "versions", "ranges") live here because they offer some context relevant to both of them.

oliverchang (commented):
thanks!

docs/schema.md
oswalpalash: Makes sense @oliverchang ! All set

oliverchang (approved):
thank you! will also let @rsc and @chrisbloom7 sign off on this.

chrisbloom7 (approved):

PR #115, opened 2023-02-22
Include last_affected in fields discussed for affected[].ranges[].type

andrewpollock:
Its omission was an oversight.


oliverchang (approved):
thanks!

PR #121, opened 2023-03-10
Clarify that the credits type field is a string.

oliverchang:
Not an array. This is how the JSON schema and examples provided encodes this.


oliverchang:
CC @hawaiigal @michaelkedar

hawaiigal (approved):

PR #122, opened 2023-03-16
Clarify schema_version being required in >1.0.0

### LINKED ISSUE #116: Can we make "schema_version" a required tag?
Can we make "schema_version" a required tag? Currently I have to brute force it by applying schemas until one passes, and it also makes identifying the files tricky.

oliverchang:
Resolves #116.

### DIFF
--- a/docs/schema.md
+++ b/docs/schema.md
@@ -182,7 +182,8 @@ RFC3339-formatted timestamp in UTC (ending in "Z"). Given two
 different entries claiming to describe the same `id` field, the one with the
 later modification time is considered authoritative.
 
-The `id` and `modified` fields are required. All other fields are optional,
+The `id` and `modified` fields are required. For schema versions above 1.0.0,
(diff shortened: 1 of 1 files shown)


kurtseifried:
It's still not required in https://github.com/ossf/osv-schema/blob/main/validation/schema.json, e.g.:

  "required": [
    "id",
    "modified"
  ]

needs to be:

  "required": [
    "id",
    "modified",
"schema_version"
  ]

oliverchang:
> It's still not required in https://github.com/ossf/osv-schema/blob/main/validation/schema.json, e.g.:
> 
> "required": [ "id", "modified" ]
> 
> needs to be:
> 
> "required": [ "id", "modified", "schema_version" ]

Unfortunately that would not be technically correct, given that a OSV 1.0 doc (without this field) should still validate with this schema.json.

chrisbloom7 (approved):
❤️

PR #131, opened 2023-03-29
Force schema_version to match the schema it came from

joshbuker:
Set to v1.4.1 as that's the next slated version.

This is mostly for completeness' sake when it comes to validating instances of OSV with things like [jsonschema](https://github.com/python-jsonschema/jsonschema). When combined with #130, it clearly defines instances of the data as following their particular semver of the schema.

### DIFF
--- a/validation/schema.json
+++ b/validation/schema.json
@@ -5,7 +5,8 @@
   "type": "object",
   "properties": {
     "schema_version": {
-      "type": "string"
+      "type": "string",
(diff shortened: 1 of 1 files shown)


joshbuker:
Closing, as this requires that OSV follows SEMVER, which isn't possible due to the compatibility requirements outlined in https://github.com/ossf/osv-schema/pull/132#issuecomment-1489414045.

Issue #90, opened 2022-10-11
Tooling for schema validation


andrewpollock:
As part of [#761](https://github.com/google/osv.dev/issues/761), we became aware that the Cloud Security Alliance has a [schema validator](https://github.com/cloudsecurityalliance/gsd-tools/tree/main/local-scripts/schema-validator).

It seems like shipping a canonical, authoritative validator tool and library with the schema would be best, rather than each ecosystem integrator needing to reinvent the wheel (and possibly less comprehensively than desired).

marco-silva0000:
I would love a canonical mypy schema object that could be imported and used in any python tooling, same for other languages.

kurtseifried:
Some comments (I'm the tool author):

We can't do this with a "pure" JSON schema because (this is the shortlist):

1. OSV only requires id and modified, so a vuln entry could basically be blank and this is "correct", I suspect the GSD might end up with a minor fork of OSV that basically just has an expanded set of required fields
2. CVE v4 has 3 schemas, one each for REJECT, RESERVED, and PUBLIC, so you need to read the state tag to know which schema to apply, CVE v5 fixes this with a single schema, you could use the JSON "one of" but it still might be incorrect
3. there are many projects using OSV, but not JSON to host the data, instead, they use YAML so either you convert the YAML to JSON, or you need a programmatic validator that can read a YAML string
4. GSD needs to implement schema validators for each data source we ingest anyways, for one simple reason: every data source we ingest typically has at least one or more errors, malformed CVE ID's, blank entries, etc. We find these and have had good luck so far getting upstreams (e.g. Mozilla, Mageia) to fix their data
5. Long term the plan is to have our validator also correct the data we serve, e.g. vendor names of "[Red Hat]", "RedHat", all should be normalized to the correct "Red Hat"

kurtseifried:
Also (obviously) we'll be releasing what we build as open source, in that directory long term so keep an eye and hopefully I'll get some more time to build this soon.

oliverchang:
+1 that a pure JSON schema is not sufficient. There are other reasons:

- We need to validate package names, versions are valid.
- We can check for e.g. `schema_version` being required when fields from a newer schema version is used. 
- And others that may overlap with @andrewpollock 's work on detecting invalid entries. Maybe some of that should be usable as a standalone tool.

marco-silva0000:
How about several different packages that all have a single responsibility?

frasertweedale:
[...]

PR #132, opened 2023-03-29
Update schema to match documentation and require schema_version

joshbuker:


oliverchang:
Hey, thanks for this change! 

However, we can't do this, because the intention of the schema validator is to provide a single schema that can validate all OSV schema versions up to the given version (Because of our backwards compatibiltiy guarantees -- see https://github.com/ossf/osv-schema/issues/116#issuecomment-1455450571).

That's not to say we shouldn't validate this, and I think we should address this as part of a tool instead, which can be a bit smarter about knowing when `schema_version` needs to be applied: https://github.com/ossf/osv-schema/issues/90

joshbuker:
Can this be distinguished with a "rolling" schema, and a versioned schema?

In v1.4.0 and above, schema_version is required. That should be reflected in the schema... If it's not, then it's not truly required.

oliverchang:
the schema_version has always been required > 1.0.0, we just haven't had the chance to implement the proposed tool to check this. 

We just made the text a bit clearer in the 1.4.0 version. 

Having two  schemas to maintain sounds like a fair bit of maintenance burden, when I think the correct approach to go with is having a smart tool that can handle all of this and more.

joshbuker:
Also, it shouldn't cause any issues for those producing osv < 1.4.0, as those versions of the schema don't require it. Only those complying to future versions would need to include it.

joshbuker:
> Having two schemas to maintain sounds like a fair bit of maintenance burden, when I think the correct approach to go with is having a smart tool that can handle all of this and more.

Using proper semver and the built-in required properties field is low maintenance and solves the issue with existing tooling like [python-jsonschema/jsonschema](https://github.com/python-jsonschema/jsonschema).

Those producing osv versions before this would be unaffected, as they would point to a version of the schema without said requirement...Unless they're pointing at main, which implying they're using the latest anyway.

oliverchang:
We've tried to avoid requiring tooling to have to understand multiple copies of the OSV schema, because that inevitably lead to a fair bit of complexity. A client consuming OSV shouldn't have to maintain several copies of the schema to validate/understand. They should just be able to rely on the latest and expect that to validate older versions as well. 

Our backwards compatibility guarantees are (from the schema documentation for `schema_version`): 

> Clients can assume that new minor and patch versions of the schema only add new fields, without changing the meaning of old fields, so that a client that knows how to read version 1.2.0 can process data identifying as schema version 1.3.0 by ignoring any unexpected fields.

Thanks for catching #139 !

joshbuker:
Would this not work by parsing >=1.0.0 && < 2.0.0?

The client parsing would be able to either pull dynamically via the canonical URL combined with the SEMVER provided in schema_version (or 1.0.0 if null), or simply use the latest schema they have compatibility with.

This is what we're doing with GSD for example, as @kurtseifried can elaborate.

PR #133, opened 2023-03-29
Add relationships field to enhance metadata on ID relationships

### LINKED ISSUE #53: Expand vuln id relationships
Allow not just aliasing (this ID is also known as...) and relates to (this ID is different from, but related to...), but also more nuanced relationships like parent/child and such.

@JLLeitschuh what would be most useful relationship types to include from a researcher's perspective?

joshbuker:
Fixes #53

### DIFF
--- a/docs/schema.md
+++ b/docs/schema.md
@@ -63,11 +63,11 @@ A JSON Schema for validation is also available
 	"withdrawn": string,
 	"aliases": [ string ],
 	"related": [ string ],
-  "relationships": [ {
-    "type": string,
-    "id": string,
-    "canonical_url": string
-  } ],
+	"relationships": [ {
+		"type": string,
+		"id": string,
+		"canonical_url": string
+	} ],
 	"summary": string,
 	"details": string,
 	"severity": [ {
--- a/docs/schema.md
+++ b/docs/schema.md
@@ -63,6 +63,11 @@ A JSON Schema for validation is also available
 	"withdrawn": string,
 	"aliases": [ string ],
 	"related": [ string ],
+  "relationships": [ {
+    "type": string,
+    "id": string,
+    "canonical_url": string
+  } ],
 	"summary": string,
 	"details": string,
 	"severity": [ {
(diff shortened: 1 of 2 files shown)


oliverchang:
Hey Josh, we can't break the schema, and would need a very good reason to create a new major version. I'll ask some more clarifying questions in #53.

joshbuker:
@oliverchang This would be backwards compatible, as you could still have an array of strings. It's just that you can now _also_ do an array of objects instead if you want. (anyOf)

joshbuker:
That said, the types I provided as examples definitely need refined and perhaps slimmed down to purely known use-cases. For now, duplicate (and perhaps alias to consolidate that?) are the only ones I know of from the GSD perspective.

@oliverchang, @darakian, are there any that would be useful from the Google/GitHub perspective? I could see GHSA's including parent/child relationships for a root-cause vulnerability and the affected OSS packages, for example.

joshbuker:
Ideally, if a v2 schema was possible:

- `related` would be removed entirely and replaced with a new `relationships` field
- `aliases` would be removed entirely and provided as a relationship type
- `relationships` would be an array of objects `{ type, id }`, similar to references being `{ type, url }`

joshbuker:
Alternatively, can also add `relationships` as a new field, and leave `aliases` and `related` unchanged. Allows for more specific relationship information, but remains completely backwards compatible and simple.

kurtseifried:
One comment: I think we should leave aliases and related as is. 1) backwards compatibility (assuming anyone uses it) and 2) it also nicely solves the case for which we have related data but don't know what the relationship is.

In my mind it would be nice to classify the relationship so that every end user doesn't have to read the links and figure it out, e.g. answers like "downstream vendor that ships this" or "vuln2 was found because of an incomplete fix for this vuln" and so on. Thus a possible solution where URLs start in related and then graduate to relationships once we know what the relationship is, also we leave the URL in related so that older tooling that can't read relationships still get the data.

joshbuker:
@darakian 

> Is that package A discovers a vuln and package B consumes A as a dependency?

This is the primary use-case I was imagining. For example, with Log4Shell you could have a root vuln for Log4j itself, with child IDs for each library or service that consumes it (Minecraft, Ubiquity, every Java application on the planet that logs things, etc). This helps keep the root vulnerability minimal and relevant while still providing the valuable meta for each package affected.

> Beyond that how do we collect and validate this data? How do we use this data? I find it very difficult to know at advisory review time what relations may exist and I'm not sure what behavior change I would expect from someone receiving an advisory if it had relationship data.

That's a good question. With something like GSD, for example, this would be folks updating the ID over time as things are discovered, rather than something frontloaded with the original/parent vuln ID.

The only behavior change I would see is some additional metadata for humans investigating the root cause. Scanners wouldn't care because it doesn't matter where the affected version comes from (one huge ID with everything everywhere affected, or from a child ID specific to the package scanned).

darakian:
> This is the primary use-case I was imagining. For example, with Log4Shell you could have a root vuln for Log4j itself, with child IDs for each library or service that consumes it (Minecraft, Ubiquity, every Java application on the planet that logs things, etc). This helps keep the root vulnerability minimal and relevant while still providing the valuable meta for each package affected.

In theory this is how CVEs are supposed to work. [Rule 7.2.4 in the cve rule set](https://www.cve.org/ResourcesSupport/AllResources/CNARules#section_7-2_how_many_vulnerabilities) states that a CNA 
> MUST NOT assign more than one CVE ID if the products are affected, because they share the vulnerable code. The assigned CVE ID will be shared by the affected products.

In practice it can be hard to know if two advisories share the same underlying vulnerable code or not which is where this gets painful.


> That's a good question. With something like GSD, for example, this would be folks updating the ID over time as things are discovered, rather than something frontloaded with the original/parent vuln ID.

> The only behavior change I would see is some additional metadata for humans investigating the root cause. Scanners wouldn't care because it doesn't matter where the affected version comes from (one huge ID with everything everywhere affected, or from a child ID specific to the package scanned).

I guess let me ask; how is a relation different from an alias? An alias is already a relation with a claim that two advisory IDs are `the same` (for some definition of sameness). I think the example you're laying out where users would tag an ID over time as new instances of a vuln are discovered could be captured in the alias field.

joshbuker:
@darakian 

> I guess let me ask; how is a relation different from an alias? An alias is already a relation with a claim that two advisory IDs are the same (for some definition of sameness). I think the example you're laying out where users would tag an ID over time as new instances of a vuln are discovered could be captured in the alias field.

An alias is a type of relationship; not all relationships are aliases. #53 has more examples of the other kinds of relationships and why they would be helpful.

darakian:
I do get wanting more fidelity out of relations, but looking at that issue it seems like an impossibility to maintain those with any sort of consistency.

docs/schema.md
```diff
@@ -238,6 +243,69 @@ same database, the duplicate entry could be written using only the `id`,
...
+  a duplicate of this ID.
+- `DUPLICATE_OF`: Points to the canonical identifier for a vulnerability within
+  a given database.
+- `INCOMPLETE_FIX_FOR`: When the remediation for a vulnerability is incomplete,
+  and causes a related vulnerability. For example, Log4Shell (CVE-2021-44228)
+  would be an incomplete fix for CVE-2021-45046.
+- `INSUFFICIENT_FIX_OF`: Fixes a vulnerability caused by a previous remediation
+  being incomplete. For example, CVE-2021-45046 would be an insufficient fix of
+  Log4Shell (CVE-2021-44228).
+- `RELATED`: An identifier that is related in an unspecified way.
```
joshbuker, line 307: @darakian Updated this with some descriptions for each relationship type.

Could you expand a bit on the concerns around consistency?

Existing tooling could continue using `aliases`/`related` or switch to `relationships` using just the `ALIAS`/`RELATED` types, so this would primarily expand what other databases would be able to do as far as data enrichment goes (one of the main use-cases for GSD).

oliverchang:
Closing. I share the same concerns with @darakian here around this being not feasible to maintain with consistency in a practical way. This also complicates the schema a fair bit, and fragments it (i.e. there are now two ways to specify aliases).

PR #134, opened 2023-03-29
Add optional schema_type for data format hinting (ala CVE)

### LINKED ISSUE #30: Please add "data_type" tag to make parsing easier
Please add "data_type" tag to make parsing easier, for example the CVE JSON format has:

 "data_type": "CVE" which means you don't have to guess and test to see what kind of data you're parsing. If the OSV ould add a " "data_type": "OSV" to everything that would make life much simpler. Especially if more people start using things like OSV's database_specific field to embed OSV data into other formats (cats and dogs living together, it's turtles all the way down).

joshbuker:
Revisiting #51

Fixes #30

### DIFF
--- a/docs/schema.md
+++ b/docs/schema.md
@@ -56,7 +56,7 @@ A JSON Schema for validation is also available
 
 ```json
 {
-  "schema_type": string,
+	"schema_type": string,
 	"schema_version": string,
 	"id": string,
 	"modified": string,
--- a/docs/schema.md
+++ b/docs/schema.md
@@ -56,6 +56,7 @@ A JSON Schema for validation is also available
 
 ```json
 {
+  "schema_type": string,
 	"schema_version": string,
 	"id": string,
 	"modified": string,
@@ -119,6 +120,18 @@ absolutely must be shared between databases, leaving customizations to the
 "ecosystem_specific" and "database_specific" blocks (see below)
 
 # Field Details
+
+## schema_type field
+
+```json
+{
+  "schema_type": "OSV"
+}
+```
+
+The `schema_type` field is a hint to parsers (and humans) that the following
+json object is explicitly OSV formatted. This string should always contain "OSV"
+when following the OSV schema.
 
 ## schema_version field
 
--- a/validation/schema.json
+++ b/validation/schema.json
@@ -4,6 +4,10 @@
   "description": "A schema for describing a vulnerability in an open source package.",
   "type": "object",
   "properties": {
+    "schema_type": {
+      "type": "string",
+      "enum": ["OSV"]
+    },
     "schema_version": {
       "type": "string"
     },


joshbuker:
This is particularly useful to distinguish things that _look_ like OSV, but are actually custom. e.g. some of the entries in ruby-advisory-db

joshbuker:
This change is fully backward compatible, as schema_type is unused historically and not required / can be ignored.

oliverchang:
Hmm, I'm not sure if we arrived at any conclusions in #51. I recall that the main point against this is that there is no standardised way to label JSON fields with their type, so for a parser to detect what format a JSON entry is, it'd need specialised knowledge of the format anyway.  If you get an OSV from somewhere (e.g. https://osv.dev, or a database export explicitly formatted as OSV), you already typically know that it's formatted as OSV. I can't think of a real use case where this would not be the case. Generally, communicating about types is also typically more the domain of the protocol serving these (e.g. mime types, file extensions).

For databases like GSD which deals with many different types of data, labelling the data can be done more consistenctly through the namespacing approach instead to wrap the various different types with the type it actually is. 

I realise that this seems like a small change to be adding, but one of the guiding principles of OSV is to be minimal, and have each field serve intentional use cases that we've encountered.

joshbuker:
This is primarily so that the data is explicit even without the context of a file server, i.e. the json/yaml can be parsed standalone without any doubts.

For example, if GSD extends the OSV schema to require fields such as summary, details, and schema_version, we would want to [use a schema_type of OSV-GSD](https://github.com/cloudsecurityalliance/gsd-tools/pull/197/files#diff-6e1711ad6a8c4ffcfef0387cee3450e66671264acc2483b7115c786dbe2a5640R9-R14) so that data could clearly be distinguished between OSV and the slightly expanded format.

This can definitely be solved with wrappers and server hinting. However in the scenario where those file servers shutdown and someone wants to do some archaeology on the archived json/yaml, this type of format hinting could be invaluable. It also reduces institutional knowledge required, by including that hinting in the data directly rather than service documentation. It also simplifies tooling, as you can dynamically scan data and deterministically validate it against the related schema (dependant of course on the schema providing declarative hinting, like [CVE](https://github.com/CVEProject/cve-schema/blob/f8f54d50eb22d94447687823c3ef1cbb518e6d86/schema/v5.0/CVE_JSON_5.0_schema.json#L351) and [CSAF](https://github.com/oasis-tcs/csaf/blob/94aa824c934fa8d63df20c7ea51d4ae25b14496d/csaf_2.0/json_schema/csaf_json_schema.json#L556) - the only other common vuln id formats - currently do).

kurtseifried:
To confirm has this been rejected?

joshbuker:
@chrisbloom7 @oliverchang Checking in on status. Is this officially rejected?

joshbuker:
> > @chrisbloom7 @oliverchang Checking in on status. Is this officially rejected?
> 
> I don't see any of the arguments against this in the original #51 addressed, so this is a no from me.

@oliverchang What are the current arguments against, just for clarity?

I feel my [earlier comment](https://github.com/ossf/osv-schema/pull/134#issuecomment-1490229467) addresses the pushback on doing this exclusively with a wrapper/server hinting, and I have yet to get feedback on why OSV can't match what the other two formats (CVE & CSAF) are currently doing when it comes to format hinting.

oliverchang:
The main point against is that this doesn't really offer any reliable way for an automated system to determine what format this is, because to make use of it you'd need to have pre-existing knowledge of the OSV format to begin with. This would ideally be addressed at a protocol level of some form. 

For your use case of identifying random files lying around without any other metadata about what they are: the same can also likely be achieved just by running the JSON validator using the latest version of the schema and also making sure that all fields exist in the schema. We can likely also improve the JSON schema / validator to make this better (e.g. setting `additionalProperties` to false among other things). 

We've also not seen any other requests for this, across all of our existing producers and consumers of OSV data. 

Adding fields to the OSV schema is very expensive, and we need to make sure every field added (even if they're optional) serves an intentional purpose.

PR #135, opened 2023-03-29
Add a type of TIMESTAMP to affected ranges

### LINKED ISSUE #25: Discussion about references field

### LINKED ISSUE #62: missing a way to specify affected/fixed version based on dates (e.g. for services)
so I found a vuln in service, they fixed it, I updated the GSD entry:

https://github.com/cloudsecurityalliance/gsd-database/pull/2366

and realized there's no way to list a date.

affected[].ranges[].type field supports SEMVER/ECOSYSTEM/GIT, and an easy fix here would be to add TIME (suggested RFC 3339 format), so can we add that, specified as RFC 3339 format? Thanks

### LINKED ISSUE #129: affected[].ranges[].type missing timestamp type
affected[].ranges[].type missing timestamp type

So we have:
https://github.com/cloudsecurityalliance/gsd-database/blob/main/2023/1001xxx/GSD-2023-1001657.json

    "product_version": "prior to Jan 4 of 2023 (2022/01/04)",
    "vulnerability_type": "XSS",
    "affected_component": "https://app.zerossl.com",

as it's a service there is no good way to specify a version. But we know for sure it was vulnerable prior to the fix time, so this is actionable information (if you have any certs from them that were created before this time you should probably roll them over).

having a TIMESTAMP type in addition to SEMVER and GIT would solve this problem easily. Can I submit a PR to add this?

As we've seen there's a lot of vulns in a lot of services and we need to start documenting them.

joshbuker:
Fixes #25
Fixes #62 
Fixes #129

### DIFF
--- a/docs/schema.md
+++ b/docs/schema.md
@@ -467,6 +467,12 @@ provides automation for auto-populating the `versions` list based on supported
 an explicitly enumerated `versions` list, because `GIT` range inclusion queries
 cannot be answered without access to a copy of the underlying Git repository.
 
+- `TIMESTAMP`: The versions `introduced` and `fixed` are RFC 3339 compliant
+  timestamps denoting a time range. This is primarily useful for humans directly
+  consuming a vulnerability to know if they were impacted by a vulnerability.
+  For example, knowing if you were impacted by CVE-2023-28858 by comparing the
+  range to when you were using the service.
+
 ### affected[].ranges[].events fields
 
 The `ranges` object's `events` field is a JSON array of objects. Each object
--- a/validation/schema.json
+++ b/validation/schema.json
@@ -117,7 +117,8 @@
                   "enum": [
                     "GIT",
                     "SEMVER",
-                    "ECOSYSTEM"
+                    "ECOSYSTEM",
+                    "TIMESTAMP"
                   ]
                 },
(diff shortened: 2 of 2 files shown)


joshbuker:

PR #136, opened 2023-03-29
Update summary and details to be required properties

joshbuker:
Having a title/description or summary/details seems like a pretty minimal and easy-to-satisfy requirement. It's also required for humans to make heads or tails of an ID they're looking at, even if it's through tooling like bundler/yarn audit.

### DIFF
--- a/validation/schema.json
+++ b/validation/schema.json
@@ -303,6 +303,8 @@
   },
   "required": [
     "id",
-    "modified"
+    "modified",
+    "summary",
+    "details"
   ]
 }


oliverchang:
This would already break some of our existing producers of data (e.g. they just provide details without summary), and I don't think we should make it harder for people to produce OSV entries. 

The intention of this minimal requirement is to allow people to reserve IDs by just producing: `schema_version: X, id: modified, last_modified: Y`.

joshbuker:
@oliverchang Counterpoint, what would be the downside of making that something along the lines of:

```json
{
  "schema_version": "x",
  "id": "x",
  "modified": "iso8601",
  "summary": "Reserved by X tooling",
  "details": "Reserved by X tooling"
}
```

Makes things explicit while still being an extremely low barrier to entry.

joshbuker:
This would however require a v2 release if semver were to be followed, and is easy enough to validate on incoming data requests.

Closing per the stance on https://github.com/ossf/osv-schema/pull/132#issuecomment-1489414045 preventing anything that would introduce an OSV v2.

PR #137, opened 2023-03-29
Add DETECTION reference type

### LINKED ISSUE #78: Additional options for references field to support reproducers/exploit code

joshbuker:
Revisiting #89 

Partially Fixes #78

Related to #138

### DIFF
--- a/docs/schema.md
+++ b/docs/schema.md
@@ -409,11 +409,11 @@ decisions about the meaning of the `ecosystem_specific` field (see below).
 
 ### affected[].severity field
 
-The `severity` field is an optional element [defined here](#severity-field). 
-This `severity` field applies to a specific package, in cases where affected 
-packages have differing severities for the same vulnerability. If any package 
-level `severity` fields are set, the top level [`severity`](#severity-field) 
-must not be set. 
+The `severity` field is an optional element [defined here](#severity-field).
+This `severity` field applies to a specific package, in cases where affected
+packages have differing severities for the same vulnerability. If any package
+level `severity` fields are set, the top level [`severity`](#severity-field)
+must not be set.
 
 ### affected[].versions field
 
@@ -714,6 +714,9 @@ The known reference `type` values are:
 
 - `ADVISORY`: A published security advisory for the vulnerability.
 - `ARTICLE`: An article or blog post describing the vulnerability.
+- `DETECTION`: A tool, script, scanner, or other mechanism that allows for
+  detection of the vulnerability in production environments. e.g. YARA rules,
+  hashes, virus signature, or other scanners.
 - `REPORT`: A report, typically on a bug or issue tracker, of the vulnerability.
 - `FIX`: A source code browser link to the fix (e.g., a GitHub commit) Note that
   the `fix` type is meant for viewing by people using web browsers.  Programs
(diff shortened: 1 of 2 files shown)


joshbuker:

docs/schema.md
```diff
@@ -714,6 +714,9 @@ The known reference `type` values are:
 
 - `ADVISORY`: A published security advisory for the vulnerability.
 - `ARTICLE`: An article or blog post describing the vulnerability.
+- `DETECTION`: A tool, script, scanner, or other mechanism that allows for
```
oliverchang, line 717: Could you please clarify what this means a bit more? 

i.e. would someone just use something like "https://www.metasploit.com/" for this? This also feels like it might be a better fit for the credits field instead?

joshbuker: Using [Log4Shell](https://github.com/cloudsecurityalliance/gsd-database/blob/main/2021/44xxx/GSD-2021-44228.json) as an example use-case for DETECTION:

- [Grep commands to detect presence of Log4Shell](https://gist.github.com/Neo23x0/e4c8b03ff8cdf1fa63b7d15db6e3860b)
- [Hashes for versions of Log4j vulnerable to Log4Shell](https://github.com/mubix/CVE-2021-44228-Log4Shell-Hashes)
- [Java pattern detection for Log4Shell](https://github.com/returntocorp/semgrep-rules/pull/1650/commits/ecfc32623eec718d61ec83b9196574f333191008/)
- [WAF Rules for kick/banning Log4Shell related strings](https://github.com/timb-machine/log4j/)
- [Burp suite scanner for Log4Shell](https://github.com/silentsignal/burp-log4shell)
- [Nmap script to detect local services for vulnerability to Log4Shell](https://github.com/Diverto/nse-log4shell)
- [Tool that generates UIDs to identify which services are vulnerable to Log4Shell](https://github.com/alexbakker/log4shell-tools)
- [Fuzzer/scanner for hosts vulnerable to Log4Shell](https://github.com/fullhunt/log4j-scan)
- [Python script to detect Log4Shell in your Java dependencies](https://github.com/takito1812/log4j-detect)

etc...

chrisbloom7: Could this not also fit under the category of `EVIDENCE` as "A demonstration of the validity of a vulnerability claim"? Not trying to be contrary, but this list could easily get unwieldy if we try to make a category for every possible type of link, so I'm trying to understand how this adds value.

joshbuker: The distinction is my mind is:

- `EVIDENCE` - Proving this vulnerability is real / exploitable in the wild, e.g. a proof-of-concept attack, or demonstration of it compromising a system
- `DETECTION` - Tooling to detect the vulnerability being exploited / affecting production environments, explicitly actionable advice for downstream / impacted users

To me, those seem worth distinguishing, if only because of the target audience for each. (`EVIDENCE` is for a maintainer to know "hey, no really, this is exploitable", while `DETECTION` is for developers/ops teams trying to mitigate the vulnerability)

chrisbloom7: Thanks for sharing your thoughts on the distinction between the two. I'm good to go if @oliverchang is.

chrisbloom7 (approved):

oliverchang (approved):

PR #138, opened 2023-03-29
Add DISCUSSION reference type

### LINKED ISSUE #78: Additional options for references field to support reproducers/exploit code

joshbuker:
Revisiting #89

Partially Fixes #78

Related to #137

### DIFF
--- a/docs/schema.md
+++ b/docs/schema.md
@@ -409,11 +409,11 @@ decisions about the meaning of the `ecosystem_specific` field (see below).
 
 ### affected[].severity field
 
-The `severity` field is an optional element [defined here](#severity-field). 
-This `severity` field applies to a specific package, in cases where affected 
-packages have differing severities for the same vulnerability. If any package 
-level `severity` fields are set, the top level [`severity`](#severity-field) 
-must not be set. 
+The `severity` field is an optional element [defined here](#severity-field).
+This `severity` field applies to a specific package, in cases where affected
+packages have differing severities for the same vulnerability. If any package
+level `severity` fields are set, the top level [`severity`](#severity-field)
+must not be set.
 
 ### affected[].versions field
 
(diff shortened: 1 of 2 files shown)


joshbuker:

docs/schema.md
```diff
@@ -409,11 +409,11 @@ decisions about the meaning of the `ecosystem_specific` field (see below).
...
-The `severity` field is an optional element [defined here](#severity-field). 
-This `severity` field applies to a specific package, in cases where affected 
-packages have differing severities for the same vulnerability. If any package 
-level `severity` fields are set, the top level [`severity`](#severity-field) 
-must not be set. 
+The `severity` field is an optional element [defined here](#severity-field).
+This `severity` field applies to a specific package, in cases where affected
+packages have differing severities for the same vulnerability. If any package
+level `severity` fields are set, the top level [`severity`](#severity-field)
+must not be set.
```
joshbuker, line 416: Looks like my IDE autostripped trailing whitespace, can undo this if necessary.

oliverchang (approved):

chrisbloom7 (commented):
`DISCUSSION` seems like it's the superset that contains `ARTICLE`, i.e. a discussion about the vulnerability. Should we deprecate `ARTICLE` in favor of this and broaden the definition of `DISCUSSION`? Otherwise it seems like there is only nuance separating `DISCUSSION` and `ARTICLE`.

joshbuker:
> `DISCUSSION` seems like it's the superset that contains `ARTICLE`, i.e. a discussion about the vulnerability. Should we deprecate `ARTICLE` in favor of this and broaden the definition of `DISCUSSION`? Otherwise it seems like there is only nuance separating `DISCUSSION` and `ARTICLE`.

I had similar thoughts, honestly. deprecated ARTICLE and incorporating it into DISCUSSION seems sane to me.

oliverchang:
deprecation doesn't mean much for the OSV schema when it comes to backwards compatibility though. I'm happy to keep this distinction (which I think there's plenty), of ARTICLE referring to blog posts and DISCUSSION referring to social media threads if you're happy with that @chrisbloom7  ?

chrisbloom7 (approved):

PR #156, opened 2023-05-25
[bitnami] Adds ecosystem and prefix for Bitnami

gongomgra:
We are in the process of creating/publishing our own CVE feed based on OSV schema for the [Bitnami catalog](https://bitnami.com/), but we would like to start the review process for a `Bitnami` ecosystem key. Do not hesitate to ask any questions you may have.

### DIFF
--- a/docs/schema.md
+++ b/docs/schema.md
@@ -287,17 +287,6 @@ The defined database prefixes and their "home" databases are:
       </td>
     </tr>
     <tr>
-      <td><code>BITNAMI</code></td>
-      <td><a href="https://github.com/bitnami/vulndb">Bitnami Vulnerability Database</a></td>
-      <td>
-        <ul>
-          <li>How to contribute: <a href="https://github.com/bitnami/vulndb/blob/main/CONTRIBUTING.md">https://github.com/bitnami/vulndb/blob/main/CONTRIBUTING.md</a></li>
-          <li>Source URL: <code>https://github.com/bitnami/vulndb/blob/main/data/&lt;component&gt;/&lt;ID&gt;.json</code></li>
-          <li>OSV Formatted URL: <code>N/A</code></li>
-        </ul>
-      </td>
-    </tr>
-    <tr>
       <td>Your database here</td>
       <td colspan="2"><a href="https://github.com/ossf/osv-schema/compare">Send us a PR</a></td>
     </tr>
--- a/docs/schema.md
+++ b/docs/schema.md
@@ -287,6 +287,17 @@ The defined database prefixes and their "home" databases are:
       </td>
     </tr>
     <tr>
+      <td><code>BIT</code></td>
+      <td><a href="https://github.com/bitnami/vulndb">Bitnami Vulnerability Database</a></td>
+      <td>
+        <ul>
+          <li>How to contribute: <a href="https://github.com/bitnami/vulndb/blob/main/CONTRIBUTING.md">https://github.com/bitnami/vulndb/blob/main/CONTRIBUTING.md</a></li>
+          <li>Source URL: <code>https://github.com/bitnami/vulndb/blob/main/data/&lt;component&gt;/&lt;ID&gt;.json</code></li>
+          <li>OSV Formatted URL: <code>N/A</code></li>
+        </ul>
+      </td>
+    </tr>
+    <tr>
       <td>Your database here</td>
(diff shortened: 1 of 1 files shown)


docs/schema.md
```diff
@@ -527,6 +538,7 @@ The defined ecosystems are:
 | `ConanCenter` | The ConanCenter ecosystem for C and C++; the `name` field is a Conan package name.  |
 | `Rocky Linux` | The Rocky Linux package ecosystem; the `name` is the name of the source package. The ecosystem string might optionally have a `:<RELEASE>` suffix to scope the package to a particular Rocky Linux release. `<RELEASE>` is a numeric version.
 | `AlmaLinux` | AlmaLinux package ecosystem; the `name` is the name of the source package. The ecosystem string might optionally have a `:<RELEASE>` suffix to scope the package to a particular AlmaLinux release. `<RELEASE>` is a numeric version.
+| `Bitnami` | Bitnami package ecosystem; the `name` is the name of the affected component. |
```
oliverchang, line 541: Is there documentation or additional rules around the naming for a "component" ?

docs/schema.md
```diff
@@ -286,6 +286,17 @@ The defined database prefixes and their "home" databases are:
         </ul>
       </td>
     </tr>
+    <tr>
+      <td><code>BITNAMI</code></td>
```
oliverchang, line 290: Do these advisories come with their own IDs? If so, we can document them under https://ossf.github.io/osv-schema/#id-modified-fields.

oliverchang (commented):
nice! Thanks for this contribution. I just have some questions.

docs/schema.md
gongomgra: We had to open a few CVEs in the past, but we don't have a specific identifier like RHSA or similar, so I will remove this from the PR. Thank you for checking!

docs/schema.md
gongomgra: We don't have any special rule for naming our components, apart from using lowercase and dashes as separators (i.e. `wordpress-multisite`). We also publish the official URL of those packages in our [Dockerfiles](https://github.com/bitnami/containers/blob/main/bitnami/wordpress/6/debian-11/Dockerfile#L23) on GitHub. 

We also provide `purl` for those packages like `pkg:bitnami/wordpress@6.2.2`.

oliverchang: Ack, thanks!

docs/schema.md
oliverchang: Apologies for my slightly confusing original comment. Nothing having a specific identifier does make things a bit more complicated within the OSV ecosystem, since if you use e.g. CVEs as the ID directly it's hard to disambiguate advisories. 

Do you have any existing examples of any Bitnami-specific advisories? It may be very worthwhile to have a specific identifier for the Bitnami-specific advisories so that it's clear exactly where it comes from and what it's specific to.

gongomgra: We only had to register CVEs for specific Bitnami products a few times in the past. And for those we filled a regular CVE on NVD. 

As we are distributing open source programs like WordPress or Redmine as provided from upstream projects (no modification to source code, just easy way to consume), we understand a custom `BITNAMI-2023-0001` ID doesn't make much sense for our specific use case. Can you help us to clarify this point?

oliverchang: To clarify, I'm suggesting that Bitnami does have its own ID namespace, to have more ownership over the Bitnami-specific bits. Even if you don't modify source code etc, there may still be interesting additional context through textual descriptions, or different severities. A cheap way to get a custom ID namespace is to just prefix CVEs with BITNAMI or something similar. e.g. BITNAMI-CVE-XXXX-XXX. 

If this doesn't make sense for you, that's OK and we can just go with CVEs. I'll merge this as-is in that case.

gongomgra: @oliverchang Thanks for your message, I really appreciate the clarification. Let me check it with the team and get back to you.

oliverchang: Yes, absolutely! Please add that ID section back to document this :)

gongomgra: Hi @oliverchang,

I have just recovered the section for `BIT` prefix instead of `BITNAMI` as we have agreed. Let me know if there is anything else I may update.

oliverchang (approved):
thanks!

Issue #150, opened 2023-04-04
Clarify use case for `last_affected` / Should `last_affected` be used for unpatched vulnerabilities?


Marcono1234:
Could you please clarify the expected use case for `last_affected`?

Currently the GitHub advisory database seems to use it (in some cases?) for unpatched vulnerabilities to describe the currently latest version when the vulnerability was discovered. Is that the intended use case, or was it rather intended for situations where the exact fix version is unknown but the vulnerability cannot be reproduced starting with a certain version, respectively can only be reproduced up to version `last_affected`?

My concern is that, especially when advisories are created without proper communication with the maintainers (which sadly still happens), new versions after `last_affected` are released which still contain the vulnerability but the advisory is not adjusted. In that case users believe they are safe even though they are not.

Examples for this usage of `last_affected` might be https://github.com/advisories/GHSA-22wj-vf5f-wrvj (and corresponding discussion in https://github.com/h2database/h2database/issues/3686) and https://github.com/advisories/GHSA-p8p7-x288-28g6. Both seem to just contain the back then latest version as `last_affected`. There are probably more cases of this, maybe even some where indeed the version after `last_affected` still contained the vulnerability, but finding such advisories might be tedious.

andrewpollock:
Hi,

We're going to review the clarity of the text at https://ossf.github.io/osv-schema/#affectedrangesevents-fields

In a nutshell, `last_affected` sets a ceiling on known to be vulnerable versions, i.e. versions "less than or equal to" this version are to be considered vulnerable and everything after this version is not vulnerable.

We've flagged this issue with GitHub to bring it to their attention. It's quite possible they're erring on the side of false negatives in favour of false positives.

rhalar:
I'm not sure I quite follow how this could be claimed, but perhaps I am not up to date in how this things get reported? In my understanding there is the situation where a vulnerability was found and the maintainer of the package issues a fix which in not part of any release (yet), but will be in the next one. In which case this is correct, the last version affected version _will_ be the one noted in `last_affected`.

But I don't believe the assumption that the next release will necessarily contain the fix holds? How then does one mark a vulnerability which has no known fix, where each subsequent release _should_ be vulnerable?
This would imply that advisories are kept up to date as new release come out, I didn't quite get the impression they are that well maintained though (understandably so I should add, this also depends on the transparency of the maintainer).

chrisbloom7:
[...]

Issue #146, opened 2023-04-04
Trying to understand why last_affected and fixed are exclusionary?


kurtseifried:
In the schema:

                {
                  "if": {
                    "properties": {
                      "events": {
                        "contains": {
                          "required": ["last_affected"]
                        }
                      }
                    }
                  },
                  "then": {
                    "not": {
                      "properties": {
                        "events": {
                          "contains": {
                            "required": ["fixed"]
                          }
                        }
                      }
                    }
                  }
                }

Why are they exclusionary? I couldn't find anything in the docs. Was there a discussion about this that I can view that explains it?

kurtseifried:
Or was it meant to have all 4 in there as per:

Only **a single type** (either `"introduced"`, `"fixed"`, `"last_affected"`,
`"limit"`) is allowed in each event object. For instance,
`{"introduced": "1.0.0", "fixed": "1.0.2"}` is **invalid**.

oliverchang:
This is this sentence in the schema:

"Entries in the events array can contain either "last_affected" or "fixed" events, but not both."

The intention is to prevent confusing combinations of last_affected and fixed within the same `events` array, because they make each other redundant.

kurtseifried:
Ok, can we update the docs to mention that this is actually enforced in the schema?

bagder:
Why can't we have both? As a mere human, I think they provide information and it certainly doesn't *hurt* to have both even if one of them would suffice.

andrewpollock:
@kurtseifried I'm trying to determine if there's anything actionable here. 

Are you asking to more explicitly state that things documented as being required by the schema are enforced by the schema definition?

kurtseifried:
> @kurtseifried I'm trying to determine if there's anything actionable here.
> 
> Are you asking to more explicitly state that things documented as being required by the schema are enforced by the schema definition?

At a minimum, it needs to be documented, and enforced in the schema, so those both need an update.

Also, I think, in line with @bagder it is not necessary to make it exclusionary especially as there are many cases where "The intention is to prevent confusing combinations of last_affected and fixed within the same events array, because they make each other redundant." is not correct (Cisco software train version, Linux Kernel, etc.).

andrewpollock:
> At a minimum, it needs to be documented, and enforced in the schema, so those both need an update.

I think the point @oliverchang was trying to make was that these are both the case today, if I am reading correctly?

Perhaps you could send us a PR to illustrate what you feel needs to be updated?

[...]

PR #159, opened 2023-06-15
Clarified last_affected

hayleycd:
This is a first pass at further clarifying the `"last_affected"` field and addressing #146 and #150. 

Preview is available [here](https://hayleycd.github.io/osv-schema/#requirements)

### DIFF
--- a/docs/schema.md
+++ b/docs/schema.md
@@ -650,10 +650,11 @@ Only **a single type** (either `"introduced"`, `"fixed"`, `"last_affected"`,
 Entries in the `events` array can contain either `"last_affected"` or `"fixed"`
 events, but not both. It's **strongly recommended** to use `"fixed"` instead of
 `"last_affected"` where possible, as it precisely identifies the version which
-contains the fix. `"last_affected"` should be thought of as a hard ceiling. 
-Versions above `"last_affected"` are considered unaffected. Unfortunately this 
-opens up the possibility for false negatives, which is why `"fixed"` is 
-overwhelmingly preferred.  
+contains the fix. `"last_affected"` should be thought of as the hard ceiling 
+of the vulnerability _at the time of publication_ in the absence of a fixed version.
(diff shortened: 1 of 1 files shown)


docs/schema.md
```diff
@@ -650,7 +650,10 @@ Only **a single type** (either `"introduced"`, `"fixed"`, `"last_affected"`,
 Entries in the `events` array can contain either `"last_affected"` or `"fixed"`
 events, but not both. It's **strongly recommended** to use `"fixed"` instead of
 `"last_affected"` where possible, as it precisely identifies the version which
-contains the fix.
+contains the fix. `"last_affected"` should be thought of as a hard ceiling. 
+Versions above `"last_affected"` are considered unaffected. Unfortunately this 
+opens up the possibility for false negatives, which is why `"fixed"` is 
+overwhelmingly preferred.
```
chrisbloom7, line 656: I think it's useful to emphasize that the use of `last_affected` is limited to specific situations where there is no known fix at the time of publication. I also would suggest slightly de-emphasizing that versions above are unaffected, as likely as that scenario is.

```suggestion
contains the fix. `"last_affected"` should be thought of as the hard ceiling 
of the vulnerability _at the time of publication_ in the absence of a fixed version.
Versions above `"last_affected"` should be considered unaffected. Unfortunately
this opens up the possibility for false negatives, which is why `"fixed"` is 
overwhelmingly preferred.
```

chrisbloom7 (approved):
Thanks for adding this clarification! I added a small suggestion based on my [response](https://github.com/ossf/osv-schema/issues/150#issuecomment-1601930082) to the originating issue

docs/schema.md
hayleycd: Agreed. Thanks!

oliverchang (approved):
Thanks @hayleycd and @chrisbloom7 for the wording fixes here!

PR #165, opened 2023-06-24
Fixed cvssv3 regex missing escape

kurtseifried:


oliverchang (approved):
thanks @kurtseifried !!

PR #166, opened 2023-06-24
Added CVSSv4 support

kurtseifried:

### DIFF
--- a/docs/schema.md
+++ b/docs/schema.md
@@ -433,6 +433,7 @@ describes the quantitative method used to calculate the associated `score`.
 | --------- | ----------- |
 | `CVSS_V2` | A CVSS vector string representing the unique characteristics and severity of the vulnerability using a version of the [Common Vulnerability Scoring System notation](https://www.first.org/cvss/v2/) that is == 2.0 (e.g.`"AV:L/AC:M/Au:N/C:N/I:P/A:C"`).|
 | `CVSS_V3` | A CVSS vector string representing the unique characteristics and severity of the vulnerability using a version of the [Common Vulnerability Scoring System notation](https://www.first.org/cvss/) that is >= 3.0 and < 4.0 (e.g.`"CVSS:3.1/AV:N/AC:H/PR:N/UI:N/S:C/C:H/I:N/A:N"`).|
+| `CVSS_V4` | A CVSS vector string representing the unique characteristics and severity of the vulnerability using a version of the [Common Vulnerability Scoring System notation](https://www.first.org/cvss/) that is >= 4.0 and < 5.0 (e.g.`"CVSS:4.0/AV:A/AC:H/AT:P/PR:L/UI:P/S:X/AU:X/R:X/V:X/RE:X/U:X/MAV:X/MAC:X/MAT:X/MPR:X/MUI:X/MVC:X/MVI:X/MVA:X/MSC:X/MSI:X/MSA:X/CR:X/IR:X/AR:X/E:X"`).|
 | Your quantitative severity type here. | [Send us a PR](https://github.com/ossf/osv-schema/compare). |
 
 ### severity[].score field
--- a/validation/schema.json
+++ b/validation/schema.json
@@ -289,7 +289,8 @@
             "type": "string",
             "enum": [
               "CVSS_V2",
-              "CVSS_V3"
+              "CVSS_V3",
+              "CVSS_V4"
             ]
           },
           "score": {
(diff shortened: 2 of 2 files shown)


kurtseifried:
https://github.com/ossf/osv-schema/issues/167

oliverchang (commented):
Thanks @kurtseifried. How stable/finalised is CVSS 4.0?

kurtseifried:
It's basically done, in theory it might change:

The CVSS v4.0 Public Preview comment period began on June 8, 2023, and runs through July 31, 2023. All feedback will be reviewed and addressed by August 31, 2023, with a target official publication date of October 1, 2023.

but in practice I think we're all pretty safe to assume this is what it is. 40+ people contributed to it representing a lot of companies, I can;t imagine someone will pop out with something major. We can also update it if needed, no?

oliverchang:
> It's basically done, in theory it might change:
> 
> The CVSS v4.0 Public Preview comment period began on June 8, 2023, and runs through July 31, 2023. All feedback will be reviewed and addressed by August 31, 2023, with a target official publication date of October 1, 2023.
> 
> but in practice I think we're all pretty safe to assume this is what it is. 40+ people contributed to it representing a lot of companies, I can;t imagine someone will pop out with something major. We can also update it if needed, no?

Do we expect any places to start publishing this before it's completely finalised? Even if the risk is small, I'd prefer that we add this to the schema once CVSS 4 is completely done to avoid potential backwards incompatible changes. More so if realistically nobody is going to use this new standard until then.

kurtseifried:
https://www.first.org/cvss/v4-0/ is now public (it was finalized some time ago).

docs/schema.md
```diff
@@ -433,6 +433,7 @@ describes the quantitative method used to calculate the associated `score`.
 | --------- | ----------- |
 | `CVSS_V2` | A CVSS vector string representing the unique characteristics and severity of the vulnerability using a version of the [Common Vulnerability Scoring System notation](https://www.first.org/cvss/v2/) that is == 2.0 (e.g.`"AV:L/AC:M/Au:N/C:N/I:P/A:C"`).|
 | `CVSS_V3` | A CVSS vector string representing the unique characteristics and severity of the vulnerability using a version of the [Common Vulnerability Scoring System notation](https://www.first.org/cvss/) that is >= 3.0 and < 4.0 (e.g.`"CVSS:3.1/AV:N/AC:H/PR:N/UI:N/S:C/C:H/I:N/A:N"`).|
+| `CVSS_V4` | A CVSS vector string representing the unique characteristics and severity of the vulnerability using a version of the [Common Vulnerability Scoring System notation](https://www.first.org/cvss/) that is >= 4.0 and < 5.0 (e.g.`"CVSS:4.0/AV:A/AC:H/AT:P/PR:L/UI:P/S:X/AU:X/R:X/V:X/RE:X/U:X/MAV:X/MAC:X/MAT:X/MPR:X/MUI:X/MVC:X/MVI:X/MVA:X/MSC:X/MSI:X/MSA:X/CR:X/IR:X/AR:X/E:X"`).|
```
oliverchang, line 436: Just a question: 

https://www.first.org/cvss/v4.0/specification-document states:

"A vector string must contain metrics in the order shown in Table 23, every other ordering is invalid. All Base metrics must be included in a vector string."

The example vector string here doesn't seem to follow the order in table 23 of the spec. e.g. "Exploit Maturity (E)" is listed in the middle of the table, but is the last metric in the example vector string.

Am I reading the spec correctly here? 

This seems different from CVSS 3.x (which states "A vector string should contain metrics in the order shown in Table 15, though other orderings are valid."). We had to remove our regex validation because  of that (https://github.com/ossf/osv-schema/issues/168).

oliverchang (commented):
Thanks @kurtseifried ! I just have some questions on this

validation/schema.json
```diff
@@ -328,6 +329,22 @@
...
+              "properties": {
+                "type": {
+                  "const": "CVSS_V4"
+                }
+              }
+            },
+            "then": {
+              "properties": {
+                "score": {
+                  "pattern": "CVSS:4\\.[0-9]\\/AV:[NALP]\\/AC:[LH]\\/AT:[NP]\\/PR:[NLH]\\/UI:[NPA](\\/VC:[HLN]\\/VI:[HLN]\\/VA:[HLN]\\/SC:[HLN]\\/SI:[HLN]\\/SA:[HLN])?\\/S:[XNP]\\/AU:[XNY]\\/R:[XAUI]\\/V:[XDC]\\/RE:[XLMH]\\/U:[XCGAR]\\/MAV:[XNALP]\\/MAC:[XLH]\\/MAT:[XNP]\\/MPR:[XNLH]\\/MUI:[XNPA]\\/MVC:[XHLN]\\/MVI:[XHLN]\\/MVA:[XHLN]\\/MSC:[XHLN]\\/MSI:[XSHLN]\\/MSA:[XSHLN]\\/CR:[XHML]\\/IR:[XHML]\\/AR:[XHML]\\/E:[XAPU]"
```
oliverchang, line 344: Same comment re ordering here. We also need to make sure the metrics listed as optional in the spec are actually optional here (per table 23 in https://www.first.org/cvss/v4.0/specification-document).

pandatix:
Hi everyone, FIRST.ORG CVSS SIG member and the man that actually did this typo after a refacto 😅 
The official Base metric ordering is `AV / AC / AT / PR / UI / VC / VI / VA / SC / SI / SA`, and will be fixed soon, so the RedHat calculator and the **official** regex provided are valid. The issue we have is that the FIRST.ORG seem to don't have an existing Errata process for the specifications so we are trying to figure out the best way to do it.

I recommend using the [official CVSS v4.0 regex in the data representations](https://www.first.org/cvss/cvss-v4.0.json) rather than a handmade one that actually does not validate string representations properly @kurtseifried.

validation/schema.json
pandatix: I recommend using the official regular expression that is fully compliant with the specification [here](https://www.first.org/cvss/cvss-v4.0.json).

docs/schema.md
pandatix: For the reason you pointed out, we had to change the fixed ordering rule. Indeed, building a CVSS v4.0-compliant regular expression imply testing O(n!) combinations, and with 32 unique metrics... The regex would be Terrabytes long 😅

The documented CVSS v4.0 vector/string representation here is invalid as the group order is BTES, not BSTE. The valid vector must be `CVSS:4.0/AV:A/AC:H/AT:P/PR:L/UI:P/VC:N/VI:N/VA:N/SC:N/SI:N/SA:N/E:X/CR:X/IR:X/AR:X/MAV:X/MAC:X/MAT:X/MPR:X/MUI:X/MVC:X/MVI:X/MVA:X/MSC:X/MSI:X/MSA:X/S:X/AU:X/R:X/V:X/RE:X/U:X`.

oliverchang:
Hey @kurtseifried ! Do you think you'll be able to find some time to address these comments? 

If you're busy, perhaps @pandatix you could help move this forward with a separate PR with the changes you think are needed?

oliverchang:
Superseded by https://github.com/ossf/osv-schema/pull/213. Thanks all!

PR #171, opened 2023-06-29
Updating gemfiles for github pages

hayleycd:
I adjusted the gemfile for the jekyll page to be more inline with the current GitHub pages documentation and to avoid a breaking change that was happening locally. 

Happy to answer any questions. 

I am running a [GitHub page](https://hayleycd.github.io/osv-schema/) from my branch, and you can inspect it if you like.


oliverchang (approved):
thanks!

PR #173, opened 2023-07-03
Add comment about `last_affected` usage in evaluation pseudocode

Marcono1234:
Resolves partially #150

I thought this might be useful, but on the other hand I am not completely sure if this is really needed and it might just increase confusion.
So no worries if you reject this PR.

### DIFF
--- a/docs/schema.md
+++ b/docs/schema.md
@@ -760,6 +760,7 @@ func IncludedInRanges(v, ranges)
         else if evt.fixed is present && v >= evt.fixed
            vulnerable = false
         else if evt.last_affected is present && v > evt.last_affected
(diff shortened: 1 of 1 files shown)


oliverchang:
Hey, thanks for this PR! https://ossf.github.io/osv-schema/#examples might be a better place for this, and it's currently lacking some examples for `last_affected`. 

CC @hayleycd who is helping us with this.

hayleycd:
I agree with @oliverchang  that this is better resolved in an example, but I appreciate your interest in the issue. 

I just opened #174 if you want to take a look.

Marcono1234:
Thanks for taking care of this and writing such an extensive example!

PR #177, opened 2023-07-17
Get GHSA importer working against with some minor tweaks

calebbrown:
- Bump `Pipfile.lock` with current dependencies (this allows installation)
- rename `osv` to `entry` to avoid shadowing the `osv` import
- add an option to pass in `--query` from the command line to manual set query parameters

### DIFF
--- a/tools/ghsa/convert_ghsa.py
+++ b/tools/ghsa/convert_ghsa.py
@@ -32,6 +32,7 @@ ECOSYSTEM_MAP = {
     'NUGET': 'NuGet',
     'COMPOSER': 'Packagist',
     'RUST': 'crates.io',
+    'ACTIONS': 'GitHub Actions',
 }
 
 NAME_NORMALIZER = {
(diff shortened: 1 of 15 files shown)


tools/ghsa/convert_ghsa.py
```diff
@@ -136,39 +137,35 @@ def convert_reference(reference: Dict[str, str]):
...
 
 
-def convert(ghsa: Dict[str, Any]):
+def convert(ghsa: Dict[str, Any]) -> Dict[str, Any]:
     """Converts a GHSA entry to an OSV entry."""
-    osv = {
-        'id':
-        ghsa['ghsaId'],
+    entry = {
+        'schema_version': '1.4.0',
```
oliverchang, line 143: nit: 1.5.0 ?

oliverchang (approved):
nice, thanks!

tools/ghsa/convert_ghsa.py
calebbrown: Fixed.

oliverchang (approved):

PR #183, opened 2023-07-19
Add a flag for removing the cvss data from the output.

calebbrown:
When dumping GHSA for Malware these values are all "0" and don't make sense.


tools/ghsa/convert_ghsa.py
```diff
@@ -185,19 +186,23 @@ def get_affected(ghsa: Dict[str, Any]) -> List[Dict[str, Any]]:
...
             name = NAME_NORMALIZER[ecosystem](name)
 
+        db = {
+            # Attribution.
+            'ghsa': ghsa['permalink'],
+            'cwes': cwes,
+        }
+
+        if include_cvss:
+            db['cvss'] = cvss
```
oliverchang, line 196: We can actulaly probably just remove these completely. The CVSS score is now part of the OSV schema is no longer necessary here.

oliverchang (approved):
LGTM to unblock

tools/ghsa/convert_ghsa.py
calebbrown: Done!

oliverchang (approved):

PR #188, opened 2023-07-28
Add SwiftURL ecosystem.

### LINKED ISSUE #170: Add SwiftRepo / SwiftURL ecosystem.
GHSA is starting to triage Swift vulnerabilities: https://github.com/advisories?query=type%3Areviewed+ecosystem%3Aswift.

Swift is a bit of a weird ecosystem (no centralized repository), so git repo  are typically the way to pull packages. This may change in the future ([proposals/0292-package-registry-service.md)](https://github.com/apple/swift-evolution/blob/main/proposals/0292-package-registry-service.md). The naming schemes are quite different so it likely makes sense to define a separate ecosystem for this in the future.

oliverchang:
Per

https://developer.apple.com/documentation/packagedescription/package/dependency
https://docs.swift.org/package-manager/PackageDescription/PackageDescription.html#package-dependency

Putting "URL" in the name to make this consistent with how it's actually defined using `Package.Dependency`. 

There are some changes coming as part of https://github.com/apple/swift-evolution/blob/main/proposals/0292-package-registry-service.md, and we'll likely need to define a new ecosystem for that once it's finalized, as it looks like the identifiers are moving to a `Scope.Name` format. 

Fixes #170.

### DIFF
--- a/docs/schema.md
+++ b/docs/schema.md
@@ -577,7 +577,7 @@ The defined ecosystems are:
 | `Photon OS` | The Photon OS package ecosystem; the `name` is the name of the RPM package. The ecosystem string must have a `:<RELEASE-NUMBER>` suffix to scope the package to a particular Photon OS release. Eg `Photon OS:3.0`. |
 | `CRAN` | The biological R package ecosystem. The `name` is an R package name. |
 | `Bioconductor` | The R package ecosystem. The `name` is an R package name. |
-| `SwiftURL` | The Swift Package Manager ecosystem. The `name` is a Git URL to the source of the package. |
+| `SwiftURL` | The Swift Package Manager ecosystem. The `name` is a Git URL to the source of the package. Versions are Git tags that comform to [SemVer 2.0](https://docs.swift.org/package-manager/PackageDescription/PackageDescription.html#version). |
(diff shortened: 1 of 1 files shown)


oliverchang:
@darakian PTAL :)

darakian:
I think we should add some language about versioning

https://docs.swift.org/package-manager/PackageDescription/PackageDescription.html#package-dependency-requirement
states
> The version rule requires Swift packages to conform to semantic versioning. To learn more about the semantic versioning standard, visit [semver.org](https://semver.org/).

My understanding is that there is no hard enforcement of that requirement
https://github.com/github/advisory-database/issues/460#issuecomment-1177169685
but perhaps stating something like `unless otherwise specified versions conform to semver 2.0` would work

darakian:
I'm not going to go as far as saying it's the ONLY reasonable course, but I've got no complaints on assuming semver 2.0 for all advisories in the ecosystem. 😄

oliverchang:
Yeah, but I do think it's the only practical approach. Without strict or well defined versioning schemes, we can't do much other than making older/non-conforming versions out of scope.

@chrisbloom7 WDYT?

chrisbloom7 (approved):
LGTM 🙇🏻

PR #213, opened 2023-11-29
Add support of CVSS v4.0

pandatix:
This PR comes to unlock #166 which seems stuck.

## What does it bring ?

Basic support of CVSS v4.0, I'll open another issue about #178 about validating the CVSS vectors soon, as the FIRST.ORG SIG CVSS provide official CVSS validation regex we could use for schema validation.

The string representation (aka vector) example comes from the CVSS v4.0 Specification Document Section 7.


pandatix:
@oliverchang here is the PR :)

oliverchang (approved):
thanks!

PR #220, opened 2024-01-15
Clarify some wording for Ubuntu ecosystem

oliverchang:


docs/schema.md
```diff
@@ -634,7 +634,7 @@ The defined ecosystems are:
...
+| `Ubuntu` | The Ubuntu package ecosystem; the `name` field is the name of the source package. The ecosystem string has a `:<RELEASE>` suffix to scope the package to a particular Ubuntu release. `<RELEASE>` is a numeric version as specified in [Ubuntu Releases](https://wiki.ubuntu.com/Releases). Also, the release version may be prefixed with `:Pro:` to denote Ubuntu Pro (aka Expanded Security Maintenance (ESM)) updates. For example, the ecosystem string "Ubuntu:22.04:LTS" refers to Ubuntu 22.04 LTS (jammy), while "Ubuntu:Pro:18.04:LTS" refers to fixes that landed in Ubuntu 18.04 LTS (bionic) under Ubuntu Pro/ESM.
```
oliverchang, line 637: question for @dodys : is ":LTS" required for specific versions? It's not mentioned in the text description.

If ":LTS" always implied by the release versions itself )i.e. we know 22.04 is LTS, 18.04 is LTS etc), could we just omit this?

dodys: The LTS is part of the full name of the release, as you can verifiy in the Releases wiki. We just didn't include the .X after a release number because those are mainly to specify point releases/ISOs.

oliverchang: Thanks for clarifying! I've updated the wording here to make ":LTS" mandatory, and to clarify that the release number should be YY.MM. How does the current wording look to you? 

Sorry for pedantic here -- it's important for us to clarify the rules so that vulnerability scanners can construct the correct names for matching.

dodys (approved):
That looks great to me, thanks!

PR #221, opened 2024-01-16
Events requirements: recommend the use of `fixed` over `limit`

### LINKED ISSUE #216: Handling of cherry picks in the specification
Currently, the specification (I'm looking at version 1.6.1, at the time of writing) does not mention how to handle [cherry pick commits](https://git-scm.com/docs/git-cherry-pick) at all. However, as I understand it, the osv.dev [implementation](https://github.com/google/osv.dev/blob/master/osv/impact.py) of the [evaluation](https://ossf.github.io/osv-schema/#evaluation) logic does handle them, to some extent.

It would be nice to specify properly how cherry picks should be handled, so that independent implementations of the evaluation logic can be created, compared, etc.

I have a few specific examples of cherry picks cases that the semantics non-trivial, which I can post here.
But before doing so I'd like to understand if there is an interest in making the OSV format specification evolve to describe how cherry picks should be handled.

zacchiro:
Rationale: reduce the amount of false negatives in commits belonging to branches forked in-between `introduced` and `fixed`/`limit`.

Closes #216.

### DIFF
--- a/docs/schema.md
+++ b/docs/schema.md
@@ -8,7 +8,7 @@ aside:
 show_edit_on_github: true
 ---
 
-**Version 1.6.2 (Jan 16, 2024)**
+**Version 1.6.3 (Jan 29, 2024)**
 
 Original authors:
 - Oliver Chang (ochang@google.com)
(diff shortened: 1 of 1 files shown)


docs/schema.md
```diff
@@ -740,6 +740,19 @@ this opens up the possibility for false negatives, which is why `fixed` is
 overwhelmingly preferred. An [example](#last_affected-vs-fixed-example) is available to
 illustrate the difference.
 
+The `fixed` and `limit` events are closely related (refer to the [Limit events
+example](#Limit-events) for further details) and involve a similar trade-off.
```
oliverchang, line 744: nit: can we qualify that this tradeoff is for "GIT" ranges only?

docs/schema.md
```diff
@@ -740,6 +740,19 @@ this opens up the possibility for false negatives, which is why `fixed` is
...
+Consider a branch forked from `main` after a vulnerability introduction
```
oliverchang, line 745: This is a lot of text to describe a situation that would likely be difficult to visualise for readers unfamiliar with these concepts. We should instead lean on the example linked to describe this. 

Would you agree with something a bit more concise, along the lines of:

The `fixed` and `limit` events are closely related and involve a similar trade-off. See the [example](#last_affected-vs-fixed-example) for an example of how they differ. `events` arrays with `fixed` events must include all other cherrypicked fix commits in all branches as separated `fixed` events to avoid false positive matches in other branches. Unlike `fixed` events, `limit` events restrict the set of vulnerable commits to ones that are reachable from the `limit`, which may result in false negatives. Where possible, it's strongly recommended to use `fixed` over `limit`.

oliverchang (commented):
super sorry for the late review!

docs/schema.md
zacchiro: Yes, sure. I've updated the PR to reflect the more concise version, with minor copy editing changes.
I've also included in it the fix for the other issue you mentioned (highlighting this is relevant for GIT ranges only).

oliverchang (approved):
Sorry for the delay here. This LGTM!

PR #229, opened 2024-03-14
Fix truncated copy/paste in database_specific description

chrisbloom7:
Adds back the `as defined by the database` in the description of the `range` object's `database_specific` field. This brings it into line with the descriptions of the other two `database_specific` fields. I suspect it was just a bad copy/paste operation as we were adding those fields to other objects beyond the top level.


oliverchang (approved):
thanks for spotting this!

PR #231, opened 2024-03-20
Enable encoding of Maven registries.

### LINKED ISSUE #208: Proposal: Supporting Maven registries
The current Maven ecosystem definition is "The Maven Java package ecosystem. The name field is a Maven package name.", which is a little vague. 

We should clarify that this is referring to Maven Central by default, and if a different registry is required, allow the ecosystem to be "Maven:<repository URL>", similar to the Linux distro definitions (e.g. "Debian:7").

The default (Maven Central) should always be used/preferred where it makes sense.

oliverchang:
Fixes #208.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,3 +29,4 @@
   definitions of `aliases` and `related`.
 - 2023-11-29 Released version 1.6.1. Some cleanup of the schema layout.
 - 2024-01-16 Released version 1.6.2. Added CVSS_V4 and Ubuntu ecosystem.
+- 2024-04-05 Released version 1.6.3. Added Maven registry support.
--- a/docs/schema.md
+++ b/docs/schema.md
@@ -8,7 +8,7 @@ aside:
 show_edit_on_github: true
 ---
 
-**Version 1.6.3 (Jan 29, 2024)**
+**Version 1.6.4 (Mar 20, 2024)**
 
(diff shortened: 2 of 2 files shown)


oliverchang:
@darakian @cuixq for review as well.

cuixq:
LGTM!

chrisbloom7 (approved):
❤️

oliverchang:
Thanks @cuixq and @chrisbloom7 :) 

@darakian do you have any comments on this? Wanted to hear your thoughts also.

darakian:
@oliverchang, sorry for the delay. My only comment is that I think it's worth adding a fully formed example for an alternate registry. Using the first artifact I see on `maven.google.com`; I believe this could be 
`Maven:maven.google.com/android.arch.core:common`

Otherwise, looks good to me 👍

oliverchang:
> @oliverchang, sorry for the delay. My only comment is that I think it's worth adding a fully formed example for an alternate registry. Using the first artifact I see on `maven.google.com`; I believe this could be `Maven:maven.google.com/android.arch.core:common`
> 
> Otherwise, looks good to me 👍

Thanks! Hm, the group ID / artifact ID is not part of the ecosystem string here. It's instead part of  the `name` field. Perhaps I could clarify that the `name` is in the `groupID:artifactID` format ?

PR #232, opened 2024-03-21
Add symlink for `schema.json` to `docs`

### LINKED ISSUE #226: Publish JSON schema to GitHub pages
As far as I can see, the JSON schema is only available at https://raw.githubusercontent.com/ossf/osv-schema/main/validation/schema.json. But `raw.githubusercontent.com` isn't really designed to be used as a CDN, which means that putting it in a `$schema` property isn't appropriate.

Copying the schema as part of the GitHub pages deployment, to something like `https://ossf.github.io/osv-schema/schema.json`, would be better in the long run

JamieMagee:
This allows the JSON schema to be published alongside the rest of the documentation to GitHub pages. The schema will be available at https://ossf.github.io/osv-schema/schema.json.

Also created a symlink to the previous repository location of `validation/schema.json` in case anyone is already using that in a `$schema` property

Closes #226

### DIFF
--- a/docs/schema.json
+++ b/docs/schema.json
@@ -1,0 +1,1 @@
+../validation/schema.json


docs/schema.json
```diff
@@ -0,0 +1,313 @@
+{
```
oliverchang, line 1: Would `validation/schema.json` being a symlink break existing users depending on the  https://raw.githubusercontent.com/ossf/osv-schema/main/validation/schema.json link ?

What if we made the docs/schema.json a symlink instead to this instead?

oliverchang (commented):
thanks!

docs/schema.json
JamieMagee: I just checked and it doesn't work 😢 https://raw.githubusercontent.com/JamieMagee/osv-schema/github-pages-schema/validation/schema.json

Unfortunately doing it the other way round doesn't work either. What do you think about duplicating the file? Or just having it in `docs`?

oliverchang: darn :/

Re "other way around doesn't work": do you mean having `validation/schema.json` be the source of truth, and having `docs/schema.json` as a symlink doesn't work either when it comes to  the GitHub site generation? 

Interestingly, it seems to work here: https://github.com/s4y/gh-pages-symlink-test

JamieMagee: I was looking through my old GitHub notifications, and just got back to this. Sorry!

I think it doesn't work for `raw.githubusercontent.com`, but does work for GitHub pages. I updated this PR to keep `validation/schema.json` and make `docs/schema.json` the symlink.

oliverchang (approved):

PR #238, opened 2024-04-30
Enhance discussion of affected[].version formats

andrewpollock:
Based on query raised in https://github.com/google/osv.dev/issues/2135

### DIFF
--- a/docs/schema.md
+++ b/docs/schema.md
@@ -678,6 +678,14 @@ The `affected` object's `versions` field is a JSON array of strings. Each string
 is a single affected version in whatever version syntax is used by the given
 package ecosystem.
 
+When there is no well-defined packaging ecosystem specified, (notably as for
+general C/C++ libraries, where there are only GIT commit ranges), this version
+format cannot be relied upon to conform to any particular syntax (e.g. they
+could be the upstream Git version tags derived from these GIT commit ranges,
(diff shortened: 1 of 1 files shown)


docs/schema.md
```diff
@@ -678,6 +678,14 @@ The `affected` object's `versions` field is a JSON array of strings. Each string
 is a single affected version in whatever version syntax is used by the given
 package ecosystem.
 
+When there is no well-defined packaging ecosystem specified, (notably as for
```
oliverchang, line 681: Hmm, there is a bit of an assumption here that it's generally accepted/understood that C/C++ libraries will only have GIT commit ranges. Would it make sense to instead say something like:

When there is no well-defined packaging ecosystem specified (for example, general C/C++ libraries), GIT commit ranges typically are the best way to define vulnerable version ranges. In this case, versions specified in this array cannot be relied upon to ...

oliverchang (commented):
thanks!

docs/schema.md
andrewpollock: Reworded in b492a29

oliverchang:
@yashrsharma44  tagging you in from https://github.com/google/osv.dev/issues/2135. Would this clarify the question you had in that?

yashrsharma44:
Thanks, this should be helpful :)

oliverchang (approved):

PR #241, opened 2024-05-13
Mention a more performant JSON validator

andrewpollock:
It does users a disservice to direct them to `check-jsonschema` in the first instance.

`check-jsonschema` is incredibly slow (probably because of Python interpreter startup time) when run in a loop on multiple files.

### DIFF
--- a/validation/README.md
+++ b/validation/README.md
@@ -5,6 +5,10 @@ This directory contains a JSON schema to validate OSV entries.
 ## Usage
 
 ```
+$ go run github.com/neilpa/yajsv@latest -s schema.json osv_to_test.json
+```
+
+```
 $ pip install check-jsonschema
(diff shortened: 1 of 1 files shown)


validation/README.md
```diff
@@ -4,6 +4,10 @@ This directory contains a JSON schema to validate OSV entries.
 
 ## Usage
 
+```
```
oliverchang, line 7: Should add a sentence that this can be used with any JSON schema validator, and that these are two examples?

andrewpollock: Clarified in [6784904](https://github.com/ossf/osv-schema/pull/241/commits/67849046d932afcd37117c0fdb6a156afec9e665)

oliverchang (approved):

PR #246, opened 2024-06-04
Validate supported ecosystems and identifier prefixes in the schema definition

andrewpollock:
Fixes: #201


validation/schema.json
```diff
@@ -280,6 +280,18 @@
     "modified"
   ],
   "$defs": {
+    "ecosystem": {
+      "type": "string",
+      "title": "Currently supported ecosystems",
+      "description": "These ecosystems are also documented at https://ossf.github.io/osv-schema/#affectedpackage-field",
+      "pattern": "^(AlmaLinux|Alpine|Android|Bioconductor|Bitnami|ConanCenter|CRAN|crates.io|Debian|GHC|GitHub Actions|GIT|Go|Hackage|Hex|Linux|Maven|npm|NuGet|OSS-Fuzz|Packagist|Photon OS|Pub|PyPI|Rocky Linux|RubyGems|SwiftURL|Ubuntu)(:[[:digit:]]+)?"
```
oliverchang, line 287: Does this regex support periods after the : bit?  e.g. `Debian:3.1` ?

andrewpollock: Any periods as per your example fall outside of the regex as currently defined, so strings as per your example pass validation:

```
$ cat /tmp/osv-vulnerabilities/Debian/DTSA-99-1.json | gron | grep ecosystem
json.affected[0].package.ecosystem = "Debian:5.0";
$ ~/go/bin/jv ~/gosst/osv/osv-schema/validation/schema.json /tmp/osv-vulnerabilities/Debian/DTSA-99-1.json
$ echo $?
0
```

oliverchang (approved):

PR #247, opened 2024-06-04
add initial reference for CGA (chainguard)

cpanato:
cc @mattmoor @luhring

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,6 +3,7 @@
 This is the repository for the Open Source Vulnerability schema (OSV Schema), which is currently exported by:
 - [AlmaLinux](https://github.com/AlmaLinux/osv-database)
 - [Bitnami Vulnerability Database](https://github.com/bitnami/vulndb)
+- [Chainguard](https://packages.cgr.dev/chainguard/osv/all.json)
 - [Curl](https://curl.se/docs/vuln.json)
 - [GitHub Security Advisories](https://github.com/github/advisory-database)
 - [Global Security Database](https://github.com/cloudsecurityalliance/gsd-database)
@@ -26,6 +27,7 @@ Together, these include vulnerabilities from:
 -   Alpine
 -   Android
 -   Bitnami
+-   Chainguard
 -   crates.io
 -   Debian GNU/Linux
 -   GitHub Actions
--- a/docs/schema.md
+++ b/docs/schema.md
@@ -388,6 +388,17 @@ The defined database prefixes and their "home" databases are:
       </td>
     </tr>
     <tr>
+      <td><code>CGA</code></td>
+      <td><a href="https://packages.cgr.dev/chainguard/osv/all.json">Chainguard Security Notices</a></td>
+      <td>
+        <ul>
+          <li>How to contribute: TBD</li>
+          <li>Source URL: TBD</li>
+          <li>OSV Formatted URL: <code>https://packages.cgr.dev/chainguard/osv/&lt;ID&gt;.json</code></li>
+        </ul>
+      </td>
+    </tr>
+    <tr>
       <td>Your database here</td>
       <td colspan="2"><a href="https://github.com/ossf/osv-schema/compare">Send us a PR</a></td>
     </tr>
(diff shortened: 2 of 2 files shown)


docs/schema.md
```diff
@@ -387,6 +387,17 @@ The defined database prefixes and their "home" databases are:
         </ul>
       </td>
     </tr>
+    <tr>
+      <td><code>CGA</code></td>
+      <td><a href="https://github.com/wolfi-dev/advisories">CGA / Wolfi - Chainguard Security Notices</a></td>
+      <td>
+        <ul>
+          <li>How to contribute: TBD</li>
```
mattmoor, line 395: I’d link to a CONTRIBUTING.md in the advisories repo, and then we can flesh that out

README.md
```diff
@@ -47,6 +48,7 @@ Together, these include vulnerabilities from:
 -   Rocky Linux
 -   RubyGems
 -   Ubuntu
+-   Chainguard
```
oliverchang, line 51: Would you mind keeping this list alphabetically sorted?

README.md
```diff
@@ -20,6 +20,7 @@ This is the repository for the Open Source Vulnerability schema (OSV Schema), wh
 - [Rust Advisory Database](https://github.com/RustSec/advisory-db)
 - [Ubuntu](https://github.com/canonical/ubuntu-security-notices/)
 - [VMWare Photon OS](https://github.com/vmware/photon/wiki/Security-Advisories) (unofficial)
+- [Chainguard](https://packages.cgr.dev/chainguard/osv/all.json)
```
oliverchang, line 23: Would you mind keeping this list alphabetically sorted?

oliverchang (commented):
Thanks @cpanato !!!! Just some comments.

docs/schema.md
```diff
@@ -636,6 +647,7 @@ The defined ecosystems are:
 | `Bioconductor` | The biological R package ecosystem. The `name` is an R package name. |
 | `Bitnami` | Bitnami package ecosystem; the `name` is the name of the affected component. |
 | `ConanCenter` | The ConanCenter ecosystem for C and C++; the `name` field is a Conan package name.  |
+| `CGA` | Chainguard package ecosystem; the `name` is the name of the affected component. |
```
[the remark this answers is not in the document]
cpanato: The dataset I've sent to start ingesting is a join of Wolfi + Chainguard, but the ecosystem is `chainguard` for all of them. 
Only `wolfi` we will work soon, and then we will add that here as well

thanks for the feedback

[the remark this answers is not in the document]
cpanato: applied some changes

docs/schema.md
```diff
@@ -636,6 +647,7 @@ The defined ecosystems are:
 | `Bioconductor` | The biological R package ecosystem. The `name` is an R package name. |
 | `Bitnami` | Bitnami package ecosystem; the `name` is the name of the affected component. |
 | `ConanCenter` | The ConanCenter ecosystem for C and C++; the `name` field is a Conan package name.  |
+| `CGA` | Chainguard package ecosystem; the `name` is the name of the package name. The package name might have a `:<RELEASE>` suffix to scope the package to a particular version stream. `<RELEASE>` is a numeric version. |
```
oliverchang, line 650: This string needs to match exactly what comes in the `package` entry. i.e. "Chainguard" here per

```
"package": {
        "ecosystem": "Chainguard",
        "name": "eks-distro-coredns-1.8",
        "purl": "pkg:apk/chainguard/eks-distro-coredns-1.8"
      },
```

Also, does Chainguard/Wolfi have a concept of source vs binary package names? If so, we should call out which this is here.

cpanato: i will confirm, but i think all is a binary package

docs/schema.md
```diff
@@ -636,6 +647,7 @@ The defined ecosystems are:
 | `Bioconductor` | The biological R package ecosystem. The `name` is an R package name. |
 | `Bitnami` | Bitnami package ecosystem; the `name` is the name of the affected component. |
 | `ConanCenter` | The ConanCenter ecosystem for C and C++; the `name` field is a Conan package name.  |
+| `CGA` | The Chainguard package ecosystem; the `name` is the name of the source package. The package name might have a `:<RELEASE>` suffix to scope the package to a particular version stream. `<RELEASE>` is a numeric version. |
```
luhring, line 650: My two cents here — 

1. We don't have source packages at this time, so there's not a valuable semantic in distinguishing these here IMHO.
2. We don't have Debian-style version streams. We do have version streams that add suffixes to the package name, but it's not meaningful for the advisory data or for vulnerability matching at scan time — the package name should be used verbatim (even if a version stream suffix is present).

```suggestion
| `CGA` | The Chainguard package ecosystem; the `name` is the name of the package. |
```

cpanato: i've add that to the commit, i did similar text before, so i think we are aligned. i was just not sure

docs/schema.md
oliverchang: Thanks. I think we still need to update the first column to be "Chainguard" instead? 

This table is referring to ecosystems, not ID prefixes.

cpanato: ah i see 🤦

cpanato: done

oliverchang (approved):

cpanato:
cool, thanks for the review @oliverchang

andrewpollock:
I note that OSV.dev is also importing [`Wolfi` ecosystem vulnerabilities](https://osv.dev/list?ecosystem=Wolfi) but it doesn't get a mention. I added validation for it in #282, but at the time didn't appreciate it wasn't in the documentation.

PR #250, opened 2024-07-10
add clarity around distros' use of aliases

luhring:
Coming out of https://github.com/google/osv.dev/issues/2374#issuecomment-2218339823, wanted to suggest some potential wording improvements to help the next Linux distro that comes along better understand how the `aliases` field should and should not be used.

I welcome any feedback, and I'm not sure I've captured the sentiment perfectly. 

One particular callout: this PR removes an existing sentence (below) that we struggled to wrap our heads around. If there's something that this was trying to convey that's lost in my PR, I'd love to better understand it.

>Aliases may be used for vulnerabilities affecting different packages or ecosystems as long as they follow this definition.

cc: @michaelkedar @andrewpollock @cpanato

### DIFF
--- a/docs/schema.md
+++ b/docs/schema.md
@@ -468,18 +468,13 @@ vulnerabilities (and no others), and vice versa.
 Aliases should be considered symmetric (if A is an alias of B, then B is an
 alias of A) and transitive (If A aliases B and B aliases C, then A aliases C).
 
-Aliases should **not** be used in records that bundle many different
-vulnerabilities in one patch of a distribution of a package. Listing multiple
-vulnerabilities as `aliases` would mean that they are all identical (due to the
-symmetry/transitivity of `aliases`), not that one release fixes multiple
-(distinct) vulnerabilities.
-
 Aliases should **not** be used to refer to vulnerabilities in packages upstream
 or downstream in a software supply chain from the given OSV record's affected
 package(s). For example, if a CVE describes a vulnerability in a language
 library, and a Linux distribution package contains that library and therefore
 publishes an advisory, the distribution's OSV record must not list the CVE ID as
-an alias. (It should use the `related` field instead.)
+an alias. Similarly, distributions often bundle multiple upstream
+vulnerabilities into a single record. `related` should be used in these cases.
 
 ## related field
 
--- a/docs/schema.md
+++ b/docs/schema.md
@@ -460,11 +460,11 @@ databases, in the form of the `id` field. This allows one database to claim that
 its own entry describes the same vulnerability as one or more entries in other
 databases.
 
-Two vulnerabilities can be described as aliases if a potential patch that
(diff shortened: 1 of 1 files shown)


cpanato (approved):
lgtm

docs/schema.md
```diff
@@ -460,11 +460,11 @@ databases, in the form of the `id` field. This allows one database to claim that
...
 databases.
 
-Two vulnerabilities can be described as aliases if a potential patch that
-addresses one of the vulnerabilities (and no other vulnerabilities) will also
-address the other vulnerability, and vice versa. Aliases may be used for
-vulnerabilities affecting different packages or ecosystems as long as they
-follow this definition.
+Two vulnerabilities can be described as aliases if they affect any given
+software component the same way: either both vulnerabilities affect the software
+component or neither do. And, consequently, a potential patch that addresses one
```
andrewpollock, line 465: Grammar nit: don't start sentences with a conjunction.

```suggestion
component or neither do. Consequently, a potential patch that addresses one
```

andrewpollock: Or more invasively, 

"A subsequent patch addresses both of the vulnerabilities (and no others), and vice versa."

docs/schema.md
```diff
@@ -475,6 +475,13 @@ vulnerabilities as `aliases` would mean that they are all identical (due to the
 symmetry/transitivity of `aliases`), not that one release fixes multiple
 (distinct) vulnerabilities.
 
+Aliases should **not** be used to refer to vulnerabilities in packages upstream
```
oliverchang, line 478: The previous point about bundling now feels a bit redundant with this definition framed around upstream/downstream. 

could we somehow merge the two? i.e. start with the definition you've provided, and then also use the bundling many different vulnerabilities bit as an additional example. Something like:

Aliases should **not** be used to refer to vulnerabilities in packages upstream 
or downstream in a software supply chain from the given OSV record's affected
package(s). For example, if a CVE describes a vulnerability in a language
library, and a Linux distribution package contains that library and therefore
publishes an advisory, the distribution's OSV record must not list the CVE ID as
an alias. Similarly, distributions often bundle multiple upstream vulnerabilities into a single record. `related` should be used in these cases. 

CC @michaelkedar for thoughts too.

docs/schema.md
luhring: I like that!

luhring: Adjusted in f84a635

docs/schema.md
luhring: I like that — adjusted in 21cbf35

oliverchang (approved):

oliverchang:
CC @chrisbloom7 and @rsc in case you have thoughts!

michaelkedar (approved):

oliverchang (approved):

PR #255, opened 2024-08-06
Update Android IDs and links in schema table.

doryiii:
Per discussion offline, this updates Android vulnerability ID prefixes to corresponds with reality. This may need to be revisited in the future (e.g. `ASB-A` -> `A-ASB`) but that would be a separate effort.

### DIFF
--- a/docs/schema.md
+++ b/docs/schema.md
@@ -157,12 +157,12 @@ The defined database prefixes and their "home" databases are:
   </thead>
   <tbody>
     <tr>
-      <td><code>A</code></td>
-      <td><a href="https://storage.googleapis.com/android-osv/">Android Vulnerability Database</a></td>
+      <td><code>`ASB-A`/`PUB-A`</code></td>
+      <td><a href="https://source.android.com/docs/security/bulletin">Android Security Bulletin</a></td>
       <td>
         <ul>
-          <li>How to contribute: TBD</li>
-          <li>Source URL: <code>N/A</code></li>
+          <li>How to contribute: <a href="https://bughunters.google.com/about/rules/android-friends/6171833274204160/android-and-google-devices-security-reward-program-rules">Android Vulnerability Rewards Program</a></li>
+          <li>Source URL: <a href="https://storage.googleapis.com/android-osv/index.html"><code>https://storage.googleapis.com/android-osv/&lt;ID&gt;.json</code></a></li>
           <li>OSV Formatted URL: <code>https://storage.googleapis.com/android-osv/&lt;ID&gt;.json</code></li>
         </ul>
       </td>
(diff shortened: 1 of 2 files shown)


docs/schema.md
```diff
@@ -157,11 +157,11 @@ The defined database prefixes and their "home" databases are:
...
     <tr>
-      <td><code>A</code></td>
-      <td><a href="https://storage.googleapis.com/android-osv/">Android Vulnerability Database</a></td>
+      <td><code>`ASB-A`/`PUB-A`</code></td>
+      <td><a href="https://source.android.com/docs/security/bulletin">Android Security Bulletin</a></td>
       <td>
         <ul>
-          <li>How to contribute: TBD</li>
+          <li>How to contribute: <a href="https://bughunters.google.com/about/rules/android-friends/6171833274204160/android-and-google-devices-security-reward-program-rules">Android Vulnerability Rewards Program</a></li>
           <li>Source URL: <code>N/A</code></li>
```
oliverchang, line 165: Can you use `https://storage.googleapis.com/android-osv/<ID>.json` here?

oliverchang (commented):
thanks!

andrewpollock:
I think the validation at https://github.com/ossf/osv-schema/blob/fca1c1348935a71bb5e93159d462b9828894c5df/validation/schema.json#L309-L309 also requires updating to reflect this.

docs/schema.md
doryiii: What do you think about putting just "https://storage.googleapis.com/android-osv" here? People can get a list of all the vulns and can drill down if they want to (we also have the "OSV Formatted URL" field right below with individual vuln URL format).

doryiii:
Updated with new source URL and updated validation json.

docs/schema.md
oliverchang: Every other source here has <ID> in it by convention, so it would be  better to be consistent here.

Ideally, this URL would be for the URL that points to some specific page on android.com for a given ID, but this doesn't exist right? (And there's also the issue with ID prefixes not matching up).

andrewpollock: Would a good compromise to be make the link text `https://storage.googleapis.com/android-osv/<ID>.json` and the hyperlink target https://storage.googleapis.com/android-osv ?

In this vein of user experience, I previously found a very nice `index.html` file (I can't remember where) that I've been sticking in our public GCS buckets to make them imminently more browse able by humans, e.g. https://storage.googleapis.com/osv-vulnerabilities/index.html (you can grab it from this bucket). It also requires an `icons` folder, which you can also grab from https://storage.googleapis.com/osv-vulnerabilities/index.html?prefix=icons/

doryiii: > make the link text https://storage.googleapis.com/android-osv/<ID>.json and the hyperlink target https://storage.googleapis.com/android-osv ?

This works.

> index.html

Thanks! We'll look into throwing this onto our bucket.

oliverchang (approved):

PR #265, opened 2024-08-29
Add UBUNTU prefix

dodys:
Also update OSV formatted URL for USN

Let me know if you think it should be `UBUNTU-CVE` instead


oliverchang (approved):
thanks!

PR #271, opened 2024-09-05
feat(suse): add `SUSE-OU-` prefixes

hogo6002:
Some optional updates (`SUSE-OU-`) also include security fixes, example: https://ftp.suse.com/pub/projects/security/osv/SUSE-OU-2015%3A1803-1.json


msmeissn (commented):
ok for me

oliverchang (approved):

PR #275, opened 2024-09-09
Fix order of prefixes in docs/schema.md

dodys:
This improves readability.

Also @oliverchang I believe CGA is missing in validation/schema.json


oliverchang (approved):
thanks!

PR #278, opened 2024-09-12
make redhat conversion tool downloadable with Pip

jasinner:
In order to install the Red Hat converter tool via Pip we need a setup.py. Adding that here.

This change avoids specifying the runtime dependencies in both Pipenv and setup.py by requesting users install the module locally with:

`pipenv install -e .`

As specified in the updated README.md


tools/redhat/setup.py
```diff
@@ -0,0 +1,28 @@
...
+
+REQUIRES = [
+    "jsonschema",
+    "requests"
+]
+
+setup(
+    name="redhat_osv",
+    version="1.0.0",
+    description="Newtopia CLI v1",
```
andrewpollock, line 15: Copy pasta?

```suggestion
    description="Convert a Red Hat CSAF document to OSV format",
```

tools/redhat/setup.py
```diff
@@ -0,0 +1,28 @@
...
+    author_email="jshepher@redhat.com",
+    url="",
+    keywords=["Swagger", "Newtopia CLI v1"],
```
andrewpollock, line 18: Maybe this instead?

```suggestion
    keywords=["OSV", "JSON", "Conversion"],
```

tools/redhat/setup.py
```diff
@@ -0,0 +1,28 @@
...
+    install_requires=REQUIRES,
+    classifiers=[
+        "Programming Language :: Python :: 3",
+    ],
+    packages=find_packages(exclude=["*_test.py"]),
+    entry_points={"console_scripts": ["convert_redhat=convert_redhat"]},
+    long_description="""\
+    The purpose of this CLI is to interact with the manifest and dependency information API from the Deptopia project, and Manifest box repo.  # noqa: E501
```
andrewpollock, line 26: Fix copy pasta

```suggestion
    The purpose of this tool is to convert from CSAF documents to OSV # noqa: E501
```

andrewpollock (changes requested):
Looks like there's a little bit of copy pasta left over?

jasinner:
Sorry about that. I've updated the value to be specific to this module as per your suggestions.

jasinner:
Justing putting this in Draft while I attempt to import it using Pip in another project.

jasinner:
> Just putting this in Draft while I attempt to import it using Pip in another project.

Ok, I had to move things around a little bit, but this is importing successfully with Pip now. @andrewpollock can you please have another look and merge if you're happy.

andrewpollock (approved):

PR #280, opened 2024-09-12
refactor(schema): simplify prefix validation regex

andrewpollock:
Annoying yak shave:

- I have an ecosystem cross-reference unit test for osv.dev in the works, which relies on using the regex in the schema definition, verbatim.

- [Python 3.7 has a FutureWarning on set constructs](https://docs.python.org/dev/whatsnew/3.7.html#re), where as far as I can tell, `(:[[:digit:]]+)?` would need to be written as `(:\[[:digit:]]+)?`, which isn't valid in the JSON Schema (more in https://github.com/python/cpython/issues/74534)

- https://json-schema.org/draft/2020-12/json-schema-core#name-regular-expressions says JSON Schema regex is ECMA-262, so what gets used in the schema needs to be the subset that works with https://docs.python.org/3/library/re.html

  - https://xkcd.com/1171/


validation/schema.json
```diff
@@ -300,7 +300,7 @@
       "type": "string",
       "title": "Currently supported ecosystems",
       "description": "These ecosystems are also documented at https://ossf.github.io/osv-schema/#affectedpackage-field",
-      "pattern": "^(AlmaLinux|Alpine|Android|Bioconductor|Bitnami|ConanCenter|CRAN|crates.io|Debian|GHC|GitHub Actions|GIT|Go|Hackage|Hex|Linux|Maven|npm|NuGet|openSUSE|OSS-Fuzz|Packagist|Photon OS|Pub|PyPI|Red Hat|Rocky Linux|RubyGems|SUSE|SwiftURL|Ubuntu)(:[[:digit:]]+)?"
+      "pattern": "^(AlmaLinux|Alpine|Android|Bioconductor|Bitnami|ConanCenter|CRAN|crates.io|Debian|GHC|GitHub Actions|GIT|Go|Hackage|Hex|Linux|Maven|npm|NuGet|openSUSE|OSS-Fuzz|Packagist|Photon OS|Pub|PyPI|Red Hat|Rocky Linux|RubyGems|SUSE|SwiftURL|Ubuntu)([0-9]+)?"
```
another-rex, line 303: would `\\d` also not work here?

oliverchang (approved):

PR #285, opened 2024-09-16
refactor(osv-linter): further tidy up output

andrewpollock:
- Use a structured error for package version findings
- tidy up the passing around of configurations
- add support for only checking specific ecosystems
Signed-off-by: Andrew Pollock <apollock@google.com>

### DIFF
--- a/tools/osv-linter/internal/checks/checks.go
+++ b/tools/osv-linter/internal/checks/checks.go
@@ -40,6 +40,7 @@ type CheckDef struct {
 // Config defines the configuration for a check.
 type Config struct {
 	Verbose bool
+	Ecosystems []string
 }
 
 // Check defines how to run the check.
--- a/tools/osv-linter/internal/checks/packages.go
+++ b/tools/osv-linter/internal/checks/packages.go
@@ -2,6 +2,7 @@ package checks
 
 import (
 	"fmt"
+	"slices"
 	"strings"
 
 	"github.com/ossf/osv-schema/linter/internal/pkgchecker"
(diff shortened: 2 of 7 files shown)


tools/osv-linter/internal/checks/checks.go
```diff
@@ -40,6 +40,7 @@ type CheckDef struct {
 // Config defines the configuration for a check.
 type Config struct {
 	Verbose bool
+	Ecosystems []string
```
oliverchang, line 43: can you add some comments to describe what these members mean exactly? It's not clear what specifying ecosystems in `Ecosystems` implies.

tools/osv-linter/internal/checks/packages.go
```diff
@@ -39,6 +40,9 @@ func PackageExists(json *gjson.Result, config *Config) (findings []CheckError) {
 		// Normalize ecosystems with a colon to their base.
 		// e.g. "Alpine:v3.5" -> "Alpine"
 		ecosystem := strings.Split(value.Get("package.ecosystem").String(), ":")[0]
+		if config.Ecosystems != nil && !slices.Contains(config.Ecosystems, ecosystem) {
```
oliverchang, line 43: I think `nil` is a valid slice value in Go? Which would imply that `slices.Contains` will just work on `nil` slice values and that we can just remove the `nil` check. 

Can you please check this?

tools/osv-linter/internal/checks/checks.go
andrewpollock: Done in 6b9abed

tools/osv-linter/internal/checks/packages.go
andrewpollock: This is an allowlist of ecosystems (when specified). Without ensuring that `config.Ecosystems` is actually set to something, `slices.Contains` does indeed return `true`, but that's not the desired behaviour. Added some explanatory comments in [5fd8b08](https://github.com/ossf/osv-schema/pull/285/commits/5fd8b080fb8b4b129abce9a89fa904bef7a2a0b2)

oliverchang: Ack

oliverchang (approved):

PR #292, opened 2024-10-09
feat: create `osvschema` go bindings package

G-Rath:
This migrates some of the "schema" based types that currently live in `osv-scanner` over to here since they're related to the underlying OSV schema rather than the scanner specifically.

Notably to help avoid dependency cycles we now have a dedicated `constants` package whose sole responsibility is to house types and constants for representing enums defined in the schema with no logic whatsoever; this will ensure that this package is always a leaf in the dependency tree, preventing circular dependencies in libraries (which notably we have now in the current `osv-scanner/models` package).

This also introduces an `ecosystem` package which houses our concept of an parsed ecosystem as implicitly defined in the spec - that is, a struct made up of an ecosystem name and an optional suffix. Note the underlying `Ecosystem` type actually lives in `constants` which might seem weird at first but 1. prevents cycles as mentioned above, and 2. avoids the weird `ecosystem.Ecosystem` situation.

Currently this is being introduced as a port from `osv-scanner` - the fact that there are missing constants will be addressed in a follow-up pull request, along with other changes such as updating the `validation/schema.json` and (hopefully) creating some automation to help keep everything in sync.


G-Rath:
@another-rex I think this is actually pretty much good to go as a package, but would appreciate you doing a check as I forgot the top level directory/package structure we actually discussed, and also not 100% confident I've setup the package correctly from a Go POV 😅

andrewpollock:
Something to consider is the developer experience around onboarding a new ecosystem, or how we'll ensure consistency and accuracy in All The Places...

Right now we're looking at:
- [`docs/schema.md`](https://github.com/ossf/osv-schema/blob/main/docs/schema.md)
- [`validation/schema.json`](https://github.com/ossf/osv-schema/blob/main/validation/schema.json)

I think if that blows out to include

- `packages/go/constants/constants.go`

we're also going to need to look at what scaffolding to put in place to ensure we maintain consistency.

G-Rath:
Yup that's exactly what I'm going to be working on next 😄

packages/go/ecosystem/ecosystem.go
```diff
@@ -0,0 +1,29 @@
+package ecosystem
+
+import (
+	"strings"
+
+	"github.com/ossf/osv-schema/constants"
+)
+
+type Parsed struct {
```
another-rex, line 9: I wonder if there's a better name for this struct. I normally would want this to just be called Ecosystem, but `ecosystem.Ecosystem` also seems a bit weird.

G-Rath: I wondered about that too but then actually really came around to it - I think it actually works really nicely with how Go recommends you do naming and that there shouldn't be any place outside of this package where the type won't be prefixed with `ecosystem`, thus it'll always be `ecosystem.Parsed`

packages/go/ecosystem/ecosystem.go
```diff
@@ -0,0 +1,29 @@
...
+	Ecosystem constants.Ecosystem
+	Suffix    string
+}
+
+func (p *Parsed) String() string {
```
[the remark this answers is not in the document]
G-Rath: ah good question - I will make sure we've got tests in place that ensure ultimately this gets JSON'd as a string rather than an object, and that it gets parsed back like that too

another-rex:
> @another-rex I think this is actually pretty much good to go as a package, but would appreciate you doing a check as I forgot the top level directory/package structure we actually discussed, and also not 100% confident I've setup the package correctly from a Go POV 😅

You might need to add another subfolder called osv-schema under go, and name the module `github.com/ossf/osv-schema/packages/go/osv-schema`, I'm not sure on that either though.

G-Rath:
> name the module github.com/ossf/osv-schema/packages/go/osv-schema

~I'm hoping not because that feels like an annoying amount of nesting, but also the linter is using `github.com/ossf/osv-schema/linter`, so at least hopefully we might be able to drop the ...~

na you must be right, cause otherwise how else could Go actually find the files? (the linter doesn't actually have a public interface, so it doesn't strictly need to follow the convention which is why it's fine)

packages/go/constants/constants.go
```diff
@@ -0,0 +1,78 @@
+package constants
+
+type Ecosystem string
+
+const (
+	EcosystemGo            Ecosystem = "Go"
```
oliverchang, line 6: nit: can we sort this alphabetically to make it easier to parse as a human?

packages/go/ecosystem/ecosystem.go
```diff
@@ -0,0 +1,64 @@
+package ecosystem
+
+import (
+	"encoding/json"
+	"strings"
+
+	"github.com/ossf/osv-schema/packages/go/constants"
```
oliverchang, line 7: nit: should we have these packages at:

`github.com/ossf/osv-schema/bindings/go/...`  instead?

packages/go/constants/constants.go
```diff
@@ -0,0 +1,78 @@
+package constants
```
oliverchang, line 1: nit: `constants` is not a great package name. How about having a top level `osv` package for things like these?

G-Rath: Given how Go naming package conventions work, I think it's actually a pretty good name given what it's trying to avoid - specifically, trying to avoid dependency cycles by having a package which should never have logic in it to avoid ever needing to do an import itself.

Having said that, `osvcontants` might be slightly better to reduce the likelihood of namespace clashes (which is the issue with naming it `osv`) 🤔

packages/go/constants/constants.go
G-Rath: yup I can do that manually, though just noting the plan is to have this automatically generated which'll take care of that too

packages/go/ecosystem/ecosystem.go
G-Rath: I think `bindings` might have been the original name we talked about, and while I don't have a strong preference either way, I think `bindings` might be a bit nicer?

packages/go/constants/constants.go
G-Rath: done - I've not sorted the others as they felt more like they'd been arranged in a natural order, but let me know if you think they should be sorted too

packages/go/ecosystem/ecosystem.go
oliverchang: @another-rex WDYT? I think it's nice to make it more explicit that these are language bindings, as opposed to arbitrary packages?

G-Rath:
I've opened #303 adding a script that handles automatically updating the constants

packages/go/ecosystem/ecosystem.go
```diff
@@ -0,0 +1,64 @@
...
+)
+
+// Parsed represents an ecosystem-with-suffix string as defined by the spec, parsed into
+// a structured format.
+//
+// The suffix is optional and is separated from the ecosystem by a colon.
+//
+// For example, "npm:abc" would be parsed into Parsed{Ecosystem: constants.EcosystemNPM, Suffix: "abc"}
```
cuixq, line 15: npm is not an ecosystem with suffix in spec. to avoid confusion, I think it's better to use some ecosystem allowing suffix.

packages/go/ecosystem/ecosystem.go
```diff
@@ -0,0 +1,64 @@
+package ecosystem
+
+import (
+	"encoding/json"
+	"strings"
+
+	"github.com/ossf/osv-schema/packages/go/constants"
+)
+
+// Parsed represents an ecosystem-with-suffix string as defined by the spec, parsed into
```
cuixq, line 10: add a link to the spec?

packages/go/ecosystem/ecosystem.go
```diff
@@ -0,0 +1,64 @@
...
+
+	if p.Suffix != "" {
+		str += ":" + p.Suffix
+	}
+
+	return str
+}
+
+// Parse parses a string into a constants.Ecosystem and an optional suffix specified with a ":"
+func Parse(str string) Parsed {
```
cuixq, line 60: since not all ecosystems allow suffix, for those that suffix is not a thing, shall we do a check and return only the ecosystem without the suffix?

cuixq: actually maybe this is beyond what does this package do - but this still be a TODO somewhere, for example validation

G-Rath: Yeah I'm of mixed feelings about that, and sadly we should decide now because we'd need to change the return to `(Parsed, error)`...

If we do want to validate it at some point, I think we should follow the `(Must)Parse` pattern, at least.

---

(I'll let others have a think and weigh in before making any changes)

packages/go/ecosystem/ecosystem.go
```diff
@@ -0,0 +1,66 @@
...
+)
+
+// Parsed represents an ecosystem-with-suffix string as defined by the [spec], parsed into
+// a structured format.
+//
+// The suffix is optional and is separated from the ecosystem by a colon.
+//
+// For example, "Debian:7" would be parsed into Parsed{Ecosystem: constants.EcosystemDebian, Suffix: "7"}
+//
+// [spec]: https://ossf.github.io/osv-schema/
```
cuixq, line 17: I think this can be more specifically pointing at https://ossf.github.io/osv-schema/#affectedpackage-field

G-Rath: I considered that, but it didn't feel right as this isn't about the `affected` field itself, it's about the "ecosystem" type/value which happens to be used by the `affected` field - I really think ultimately the ecosystems table should have its own dedicated header as it's a very important part of the doc yet you currently can't directly link to it

cuixq: I don't see issue making the ecosystems table to have its own link. @oliverchang what do you think?

another-rex: yeah, I also got a few requests from others to make it easier to link to the ecosystems table

G-Rath: I'll have a go at a PR for that then

G-Rath: https://github.com/ossf/osv-schema/pull/306

packages/go/ecosystem/ecosystem.go
another-rex: Hmm I can see the benefit of knowing Parsed contents will always be valid OSV-Schema Ecosystem, but at the same time I think that's a decent bit of logic we need to add here, and makes it less forward compatible (if you are using an older version of osv-schema golang package, you'll get errors when trying to parse new ecosystems). 

I think let's keep it "dumb" and just do a simple cut and type cast like it is currently doing.

packages/go/constants/constants.go
oliverchang: How about `osvschema` ? We would keep it as a place for defining structs and constants, but it would have a slightly more descriptive name.

oliverchang: @another-rex

G-Rath: Sure, just so long as it's clear when you say "structs", that the idea is there should never be any logic.

The whole idea around this was it'd be a constant-only package to ensure there will never be any cycles - while in theory it can have structs too, imo that makes it very tempting to have logic. We can try to avoid that but it's required for any special handling of marshalling, and then that makes it tempting to have _just a bit more logic_, and then we're back where we are today with cyclical dependencies...

That's why we decided we'd have this package dedicated to just constants, and have all the structs in another package

oliverchang: By struct, I mean any structs for e.g. defining the OSV schema structures. 

As a whole, I don't think any bindings code in this repository should have much logic at all on top of just defining constants and structs. It should just contain a bunch of definitions + some very lightweight helpers to help others use the OSV format. 

We should keep things to a bare minimum, and maintain more complicated logic elsewhere.

packages/go/ecosystem/ecosystem.go
oliverchang: +1 to keeping it dumb.

packages/go/constants/constants.go
another-rex: happy to go with `osvschema`

packages/go/ecosystem/ecosystem.go
another-rex: we need to repeat the word `osv-schema` either way I think, otherwise the package would still be called `go`.

happy to have bindings as an intermediate directory.

packages/go/constants/constants.go
G-Rath: done

packages/go/ecosystem/ecosystem.go
G-Rath: done

another-rex:
I talked with @oliverchang offline and I think we should remove Parsed for now from the osv-schema repository so that this repository only contains types, constants, and structs with 0 code.

 (and probably move Parsed into osv-scanner instead, but that can be a later decision).

bindings/go/ecosystem/ecosystem.go
```diff
@@ -0,0 +1,66 @@
+package ecosystem
```
oliverchang, line 1: Let's exclude this for now, since it's not clear this should belong in the schema repository.

G-Rath:
Ok so have removed the `ecosystem` package, and for now `osvschema` has just constants - I'll do another pull request exploring add structs, since they can be more complex and the ones we've got in `osv-scanner` do have some logic for marshalling JSON and YAML

oliverchang (approved):

PR #294, opened 2024-10-15
strip arch from version and deduplicate arch agnostic packages

jasinner:
Including the arch suffix in a fixed version can cause false positives if the wrong arch range is used to scan a system.

For example

`0:7.3.0-427.28.1.el9_4.aarch64` < `0:7.3.0-427.28.1.el9_4.ppc64le`

So `0:7.3.0-427.28.1.el9_4.aarch64` would match the `0..0:7.3.0-427.28.1.el9_4.ppc64le` range and be considered vulnerable

### DIFF
--- a/tools/redhat/redhat_osv/osv.py
+++ b/tools/redhat/redhat_osv/osv.py
@@ -111,14 +111,8 @@ class Affected:
     Class to hold affected data for a Vulnerability
     """
 
-    remediation: InitVar[Remediation]
-    package: Package = field(init=False)
-    ranges: list[Range] = field(init=False)
-
-    def __post_init__(self, remediation: Remediation):
-        self.package = Package(remediation.component, remediation.cpe,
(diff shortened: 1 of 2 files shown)


andrewpollock (commented):
Does this have the (unintended?) consequence of making it impossible to reason about source-versus-binary RPM vulnerabilities?

andrewpollock:
> Does this have the (unintended?) consequence of making it impossible to reason about source-versus-binary RPM vulnerabilities?

Discussed in person. Decided to cross this bridge if and when a concrete use case for source-versus-binary package scanning presents itself. Realistically, this shouldn't be expressed in the version string.

jasinner:
Putting this in draft while I test for any runtime errors on the CSAF data set.

andrewpollock (approved):

abdosalamaljanad:
Thank you []()

PR #295, opened 2024-10-18
feat(linter): add a check for affected field

hogo6002:
`affected:` should be defined in the data source and should not be an empty array.


tools/osv-linter/internal/checks/affected.go
```diff
@@ -0,0 +1,21 @@
+package checks
+
+import (
+	"github.com/tidwall/gjson"
+)
+
+var CheckAffectedFieldValid = &CheckDef{
+	Code:        "A0001",
+	Name:        "affected-field-valid",
+	Description: "affected field validates",
```
andrewpollock, line 10: The style I've been going with is more like a test assertion

```suggestion
	Description: "every record has affected data",
```

tools/osv-linter/internal/checks/affected.go
```diff
@@ -0,0 +1,21 @@
+package checks
+
+import (
+	"github.com/tidwall/gjson"
+)
+
+var CheckAffectedFieldValid = &CheckDef{
+	Code:        "A0001",
+	Name:        "affected-field-valid",
```
andrewpollock, line 9: This names it more like the existing `introduced-event-exists` check"

```suggestion
	Name:        "affected-data-exists",
```

tools/osv-linter/internal/checks/affected.go
```diff
@@ -0,0 +1,21 @@
...
+	Description: "affected field validates",
+	Check:       AffectedFieldValid,
```
andrewpollock, line 11: Naming nit.

```suggestion
	Check:       RecordHasAffected,
```

tools/osv-linter/internal/checks/affected.go
```diff
@@ -0,0 +1,21 @@
...
+}
+
+// AffectedFieldValid checks if the 'affected' field exists in the JSON and is not an empty array.
+func AffectedFieldValid(json *gjson.Result, config *Config) (findings []CheckError) {
```
andrewpollock, line 15: Naming nit.

```suggestion
// RecordHasAffected checks if the 'affected' field exists in the JSON and is not an empty array.
func RecordHasAffected(json *gjson.Result, config *Config) (findings []CheckError) {
```

tools/osv-linter/internal/checks/affected.go
```diff
@@ -0,0 +1,21 @@
+package checks
+
+import (
+	"github.com/tidwall/gjson"
+)
+
+var CheckAffectedFieldValid = &CheckDef{
+	Code:        "A0001",
```
andrewpollock, line 8: (Over) thinking about codes, and prefixes, IMO, this is a record-level check, but I'm already using "R" for ranges...

andrewpollock (commented):
This LGTM as is, all of my feedback is on naming.

tools/osv-linter/internal/checks/affected.go
hogo6002: I also feel I shouldn't name this file as `affected.go`?. `record.go` sounds better and can add more checks in the future.

hogo6002: I renamed the file to `record.go` and renamed the check to `CheckRecordHasAffected`, but i didn't change the code. I agree that this is a record-level check and "R" is better than "A". Should we rename the range code to "V" for versions?

andrewpollock: > Should we rename the range code to "V" for versions?

Versions are one of the types of things in ranges :-)

Maybe for record-level, we go with the very non-obvious "O" for OSV? Naming is hard...

andrewpollock (approved):
LGTM, this is good enough for now. We can worry about improving the code later (famous last words)

PR #296, opened 2024-10-20
feat: create dedicated ecosystem name enum with all ecosystems

G-Rath:
This introduces new dedicated types for the name of an ecosystem (which is an enum) and an ecosystem suffix (which is a string starting with `:`), along with renaming the existing `ecosystem` type to `ecosystemWithSuffix` (+ adding it).

While these types are not actually used in the schema due to JSONSchema currently does not have a way of specifying that a string is made up of two types, having them defined enables a lot of automation as the current `pattern` based type cannot be used to get a finite list of all possible ecosystems defined by the spec.

As an example, here's a [TypeScript playground](https://www.typescriptlang.org/play/?#code/C4TwDgpgBAogxgewM4icCBbKBeKByOAJwEN0kA6ASwTygB98BxG+-AWWIDcIA7PAWABQQ0JFiIUaTAHVKwABYAFQhABmlAB45xyVOiwMABgBIA3vF1SMAXwBcZtIUo8A5tYfAnr64YDcQoQATCDgAG2IVKFUAVx44YGoeKJ4ACg1bHUl9WQVlNU0ASgzOBEpA-2FBVVSCEjIqBFsARgAmPAKhapTa0ggKanbOmqJe-sbByud0QlViOGhFCKQIQIsszChTISgoEMt9DLW9TAqdpGjVdXSoR2cXCusgA) showcasing how you could use the types that would be automatically generated from this new schema structure using [`json-schema-to-typescript`](https://www.npmjs.com/package/json-schema-to-typescript) (with a manually defined `EcosystemSuffix` type, though we can later introduce that for TypeScript specifically by adding a `tsType` property).

Notably, landing this will unblock me adding automation to ensure that all lists of the ecosystem name are in sync (notably, the table in the docs, the enum + pattern here, and soon the constants being added in #292)

### DIFF
--- a/validation/schema.json
+++ b/validation/schema.json
@@ -56,7 +56,7 @@
             "type": "object",
             "properties": {
               "ecosystem": {
-                "$ref": "#/$defs/ecosystem"
+                "$ref": "#/$defs/ecosystemWithSuffix"
               },
               "name": {
                 "type": "string"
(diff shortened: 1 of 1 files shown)


validation/schema.json
```diff
@@ -296,11 +296,55 @@
     "modified"
   ],
   "$defs": {
-    "ecosystem": {
```
another-rex, line 299: I wonder if changing this ecosystem name will break anyone's workflow.

another-rex (approved):
LGTM, fyi @andrewpollock & @oliverchang

validation/schema.json
G-Rath: I thought about that, but figured the likelihood anyone is depending on the name specifically right now is pretty low especially given no one seems to have caught how out of date this field actually is 😅

(the name should only matter if you're extending the schema in some way - if you're just using it for validating JSON files then at most it should result in a slightly different error message which shouldn't be depended on anyway)

G-Rath:
I've opened #303 adding a script that handles automatically updating the enum and pattern

oliverchang (approved):

validation/schema.json
andrewpollock: Ladies and gentlemen, I present to you https://github.com/google/osv.dev/blob/aa4785301751c2ba229b07dd01d9dd685721c388/osv/ecosystems/ecosystems_test.py#L28-L32, which did indeed break :-)

No biggie, and I didn't pick up on this at the time...

PR #298, opened 2024-10-21
ci: update `tj-actions/changed-files`

G-Rath:
This is the current latest version, with none of the breaking changes being impactful though [v44.0.0's is noteworthy for forks](https://github.com/tj-actions/changed-files/releases/tag/v44.0.0)


G-Rath:
(the main reason for doing this is it should be straightforward to land, meaning I'll then be a contributor so shouldn't need workflow runs approved 😅)

andrewpollock (approved):

PR #300, opened 2024-10-21
Clarify PURL requirements.

oliverchang:
PURLs should not include the `@version` component when used in OSV. 

`affected[].ranges[]`  should be used for this purpose.


andrewpollock (approved):
LGTM. Do you want to explicitly state anything about the other components?

oliverchang:
> LGTM. Do you want to explicitly state anything about the other components?

I'm not sure we have much to add there. Are there any qualifiers in particular you'd want to see not be included in PURLs here?

chrisbloom7 (approved):

PR #303, opened 2024-10-21
feat: create script for updating lists of ecosystems

G-Rath:
This introduces a new script to make it easier to ensure all "lists of ecosystems" within this codebase remain up to date, including:
  - the table in `docs/schema.md` 
  - the Go constants being introduced in #292
  - the JSON schema in `validation/schema.json` (both the pattern and the enum being introduced in #296)

To make it a bit easier, I've introduced a top-level `ecosystems.json` which is a map of defined ecosystems and a markdown description, sorted alphabetically (which the script also ensures) - I felt this was easier than trying to extract the list from markdown or another source, though it does mean double quotes need to be manually escaped.

I went with JSON as it can be read without requiring an external dependency, though if we use Python 3.11 we could switch to `toml` instead as that ships with `tomllib`

Example of the workflow output:

image

image


G-Rath:
https://github.com/ossf/osv-schema/pull/292#discussion_r1811556242 made me think we should later expand `ecosystems.json` to include other properties like if an ecosystem supports a prefix i.e.

```
{
  "npm": {
    "description": "xyz",
    "supportsSuffix": false
  }
}
```

validation/schema.json
G-Rath, line 1: note: a large amount of the changes here are because a section of the schema is incorrectly indented

I've opened #307 to reformat the file independently to remove this

docs/schema.md
```diff
@@ -683,46 +683,49 @@ versions and different potential vulnerabilities:
 
 #### Defined ecosystems
 
-<!-- Please keep this list alphabetically sorted -->
 The defined ecosystems are:
 
-| Ecosystem | Description  |
-| --------- |-----------------|
-| `AlmaLinux` | AlmaLinux package ecosystem; the `name` is the name of the source package. The ecosystem string might optionally have a `:<RELEASE>` suffix to scope the package to a particular AlmaLinux release. `<RELEASE>` is a numeric version.
+<!-- begin auto-generated ecosystems list -->
```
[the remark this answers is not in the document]
G-Rath: I've added a little something - I don't think it's perfect but hopefully enough to unblock for now

(I imagine we can improve on this over time as we see how our usage of `ecosystems.json` and co evolve)

scripts/update-ecosystems-lists.py
```diff
@@ -0,0 +1,144 @@
...
+
+  names = ecosystems.keys()
+  pattern = ""
+
+  for name in names:
+    pattern += name.replace(".", "\\.")
+    pattern += "|"
+
+  # this is a special "ecosystem" name
+  pattern += "GIT"
```
oliverchang, line 35: nit: single quotes for consistency.

oliverchang (approved):

oliverchang (approved):

PR #306, opened 2024-10-23
feat: add dedicated header for ecosystems table

G-Rath:
The table is pretty important but currently you can't actually link directly to it, so this adds a header for that


G-Rath:
Technically this works, though it feels slightly weird having it bang in the middle of the field definitions - I feel like ideally it maybe should be its own "ecosystems" appendix, but maybe that's too much..?

oliverchang (approved):

PR #308, opened 2024-11-04
fix(redhat_osv): restructure test data

andrewpollock:
This commit restructures the test data so that CSAF and OSV data is not co-mingled in the same directory. It is an attept to address

```
$ go install github.com/ossf/osv-schema/tools/osv-linter@latest
go: downloading github.com/ossf/osv-schema v1.6.7
go: github.com/ossf/osv-schema/tools/osv-linter@latest: create zip: tools/redhat/testdata/rhsa-2024_4546.json: case-insensitive file name collision: "tools/redhat/testdata/RHSA-2024_4546.json" and "tools/redhat/testdata/rhsa-2024_4546.json"
```


andrewpollock:
@jasinner FYI, also what's the go with `"test_date"` here? https://github.com/ossf/osv-schema/blob/34c8f0fb9794582b655902acd47dc6ef7db6e0b8/tools/redhat/redhat_osv/osv_test.py#L30-L30

jasinner:
Looks good to me. If we wanted to match more closely with the filenames produced by the redhat_osv converter we could change the expected output filename to use a colon like 'RHSA-2024:6220.json'. I just didn't do that because the colon character needs to be escaped in a Linux shell.

jasinner:
> test_date



> what's the go with `"test_date"` here?

The short version is that the date is not used in the test. The long version is that the 'modified' date used in the OSV class is passed into the converter instead of being calculated based on the current time. We we do the conversion of the data we use the same modified date as the CSAF file because we that value to determine if the content has changed or not since the last update.

This has the unfortunate side affect that if the OSV data is regenerated with new conversion code the modified date of the OSV records will not be updated.

oliverchang (approved):

PR #310, opened 2024-11-07
Clarify sentence about "last_affected" and "fixed" in docs.

tiegz:
From the docs:

> Entries in the events array can contain either "last_affected" or "fixed" events, but not both.

This line could be referring to either of these events arrays:

``` json
{
  "events": [
    {"introduced": "0.0.0"},
    {"last_affected": "1.0.0", "fixed": "1.0.1"}
  ]
}
```

``` json
{
  "events": [
    {"introduced": "0.0.0"},
    {"last_affected": "1.0.0"},
    {"fixed": "1.0.1"}
  ]
}
```

The first one is invalid because it has an Object that contains two entries, but that is covered by this line from the docs:

> Only a single type (either introduced, fixed, last_affected, limit) is allowed in each event object.

And according to this comment I think the **latter** invalid case is the one that is being referred to: https://github.com/ossf/osv-schema/issues/146#issuecomment-1495436723

So I suggest changing the wording to be more specific:

> An events array can have entries containing either "last_affected" or "fixed" events, but not both.

### DIFF
--- a/docs/schema.md
+++ b/docs/schema.md
@@ -1058,7 +1058,7 @@ Only **a single type** (either `introduced`, `fixed`, `last_affected`,
 `limit`) is allowed in each event object. For instance,
 `{"introduced": "1.0.0", "fixed": "1.0.2"}` is **invalid**.
 
-An events array can have entries containing either "last_affected" or "fixed" events, 
+Entries in the `events` array may be "last_affected" or "fixed" events, 
 but not both. It's **strongly recommended** to use `fixed` instead of
 `last_affected` where possible, as it precisely identifies the version which
 contains the fix. `last_affected` should be thought of as the hard ceiling


docs/schema.md
```diff
@@ -830,8 +830,8 @@ Only **a single type** (either `introduced`, `fixed`, `last_affected`,
 `limit`) is allowed in each event object. For instance,
 `{"introduced": "1.0.0", "fixed": "1.0.2"}` is **invalid**.
 
-Entries in the `events` array can contain either `last_affected` or `fixed`
-events, but not both. It's **strongly recommended** to use `fixed` instead of
+An events array can have entries containing either "last_affected" or "fixed" events,
```
andrewpollock, line 833: Yes, there's some semantic nuance here, which is a little tricky to convey, I think.

So to get all JSON-technical, the `events` array is an array of JSON (event) _objects_, and the intention of the statement under scrutiny is to refer to these event objects, as you've surmised.

https://github.com/ossf/osv-schema/blob/main/validation/schema.json#L92-L150

```suggestion
Entries in the `events` array may be "last_affected" or "fixed" events, 
```

tiegz: @andrewpollock I think that it's possible that someone may still misread that suggestion as disallowing this case:

```
[..., {"last_affected": "1.0.0", "fixed": "1.0.1"}, ...]
```

andrewpollock: the "but not both" still remains, so that should be pretty clear?

another-rex (approved):

another-rex:
Looks good, though can you have a look at the DCO check and fix it before it can be merged. Thanks!

PR #313, opened 2024-11-19
swich CSAF type from csaf_vex to csaf_security_advisory

jasinner:
Red Hat changed the document.category on 31st Oct. Therefore any records since then have not been processed. 

See https://access.redhat.com/articles/5554431


oliverchang (approved):
thanks!

PR #317, opened 2024-12-03
docs(schema): clarify purls must not contain versions

andrewpollock:
This clarifies the intent of the optional `purl` field to refer to the package in an analogous way to the `ecosystem` and `name` and not contain any explicit `version` component.

### DIFF
--- a/docs/schema.md
+++ b/docs/schema.md
@@ -668,7 +668,8 @@ within its ecosystem. The two fields must both be present, because the
 The `purl` field is a string following the
  [Package URL specification](https://github.com/package-url/purl-spec) that
 identifies the package, without the `@version` component.
-This field is optional but recommended.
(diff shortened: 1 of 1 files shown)


docs/schema.md
```diff
@@ -668,7 +668,8 @@ within its ecosystem. The two fields must both be present, because the
 The `purl` field is a string following the
  [Package URL specification](https://github.com/package-url/purl-spec) that
 identifies the package, without the `@version` component.
-This field is optional but recommended.
+This field is optional but recommended. It *must not* contain a `version`
```
oliverchang, line 671: this is already made explicit on line 670?

andrewpollock: Huh. Couldn't see it for the trees at the time :-(

PR #319, opened 2024-12-04
Add Kubernetes ecosystem

knqyf263:
Adding `Kubernetes` ecosystem according to the discussion [here](https://github.com/kubernetes-sigs/cve-feed-osv/issues/9).

This PR does not specify how to handle Kubernetes distributions by cloud vendors like EKS and GKE. We can talk about it here if needed, or merge it once and discuss it separately.

### DIFF
--- a/docs/schema.md
+++ b/docs/schema.md
@@ -8,7 +8,7 @@ aside:
 show_edit_on_github: true
 ---
 
-**Version 1.6.7 (Sep 16, 2024)**
+**Version 1.6.8 (Dec 4, 2024)**
 
 Original authors:
 - Oliver Chang (ochang@google.com)
@@ -285,6 +285,17 @@ The defined database prefixes and their "home" databases are:
           <li>How to contribute: <a href="https://github.com/haskell/security-advisories/blob/main/CONTRIBUTING.md">https://github.com/haskell/security-advisories/blob/main/CONTRIBUTING.md</a></li>
           <li>Source URL: <code>TBD</code></li>
           <li>OSV Formatted URL: <code>https://raw.githubusercontent.com/haskell/security-advisories/main/advisories/&lt;ID&gt;.json</code></li>
+        </ul>
+      </td>
+    </tr>
+    <tr>
+      <td><code>KUBE</code></td>
+      <td><a href="https://github.com/kubernetes-sigs/cve-feed-osv">Kubernetes Official CVE Feed</a></td>
+      <td>
+        <ul>
+          <li>How to contribute: <a href="https://github.com/kubernetes-sigs/cve-feed-osv/blob/main/CONTRIBUTING.md">https://github.com/kubernetes-sigs/cve-feed-osv/blob/main/CONTRIBUTING.md</a></li>
+          <li>Source URL: <code>https://kubernetes.io/docs/reference/issues-security/official-cve-feed/index.json</code></li>
+          <li>OSV Formatted URL: <code>https://raw.githubusercontent.com/kubernetes-sigs/cve-feed-osv/blob/main/vulns/&lt;ID&gt;.json</code></li>
         </ul>
       </td>
     </tr>
@@ -706,6 +717,7 @@ The defined ecosystems are:
 | `Go` | The Go ecosystem; the `name` field is a Go module path. |
 | `Hackage` | The Haskell package ecosystem. The `name` field is a Haskell package name as published on Hackage. |
 | `Hex` | The package manager for the Erlang ecosystem; the `name` is a Hex package name. |
+| `Kubernetes` | The Kubernetes ecosystem; the `name` field is a Kubernetes component name. |
 | `Linux` | The Linux kernel. The only supported `name` is `Kernel`. |
 | `Mageia` | The Mageia Linux package ecosystem; the `name` is the name of the source package. The ecosystem string must have a `:<RELEASE-NUMBER>` suffix to scope the package to a particular Mageia release. Eg `Mageia:9`. |
 | `Maven` | The Maven Java package ecosystem. The `name` field is a Maven package name in the format `groupId:artifactId`. The ecosystem string might optionally have a `:<REMOTE-REPO-URL>` suffix to denote the remote repository URL that best represents the source of truth for this package, without a trailing slash (e.g. `Maven:https://maven.google.com`). If this is omitted, this is assumed to be the Maven Central repository (`https://repo.maven.apache.org/maven2`). |
--- a/docs/schema.md
+++ b/docs/schema.md
@@ -717,7 +717,7 @@ The defined ecosystems are:
 | `Go` | The Go ecosystem; the `name` field is a Go module path. |
 | `Hackage` | The Haskell package ecosystem. The `name` field is a Haskell package name as published on Hackage. |
 | `Hex` | The package manager for the Erlang ecosystem; the `name` is a Hex package name. |
-| `Kubernetes` | The Kubernetes ecosystem; If the component is a Go module, then the `name` filed is the Go module name. |
+| `Kubernetes` | The Kubernetes ecosystem; `name` is the Go module name associated with the relevant Kubernetes component (e.g. `k8s.io/apiserver`) |
 | `Linux` | The Linux kernel. The only supported `name` is `Kernel`. |
 | `Mageia` | The Mageia Linux package ecosystem; the `name` is the name of the source package. The ecosystem string must have a `:<RELEASE-NUMBER>` suffix to scope the package to a particular Mageia release. Eg `Mageia:9`. |
 | `Maven` | The Maven Java package ecosystem. The `name` field is a Maven package name in the format `groupId:artifactId`. The ecosystem string might optionally have a `:<REMOTE-REPO-URL>` suffix to denote the remote repository URL that best represents the source of truth for this package, without a trailing slash (e.g. `Maven:https://maven.google.com`). If this is omitted, this is assumed to be the Maven Central repository (`https://repo.maven.apache.org/maven2`). |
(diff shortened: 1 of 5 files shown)


docs/schema.md
```diff
@@ -706,6 +717,7 @@ The defined ecosystems are:
 | `Go` | The Go ecosystem; the `name` field is a Go module path. |
 | `Hackage` | The Haskell package ecosystem. The `name` field is a Haskell package name as published on Hackage. |
 | `Hex` | The package manager for the Erlang ecosystem; the `name` is a Hex package name. |
+| `Kubernetes` | The Kubernetes ecosystem; the `name` field is a Kubernetes component name. |
```
oliverchang, line 720: Is there some documentation we can point to as an authoritative reference for what a component name is and how to format it?

oliverchang (commented):
thank you so much for this PR @knqyf263 !

docs/schema.md
oliverchang: To clarify this comment a bit more using your example in https://github.com/kubernetes-sigs/cve-feed-osv/issues/9#issuecomment-2506104722:

It's unclear if the name should be "kube-apiserver", or "k8s.io/apiserver". Is there some clarification we can make here on the expected naming scheme?

oliverchang (commented):
happy new year @knqyf263 ! 

Just following up on your PR once more :)

docs/schema.md
knqyf263: Hello @oliverchang, sorry for the late reply. I had been on vacation, and I've missed your comment in accumulated tasks.

Actually, I would like to discuss this point: I thought it would be difficult to distinguish `apiserver` from non-upstream apiserver if we use `apiserver`, so I adapted it to Go's module name, like `k8s.io/apiserver` to clarify it's upstream, but it is not yet strictly defined. How can we distinguish between different flavors within the same ecosystem in the OSV schema?

oliverchang: We have the convention of having a ":SUFFIX" qualifier for ecosystem names. 

For example, for [Debian](https://ossf.github.io/osv-schema/#affectedpackage-field):

> The Debian package ecosystem; the name is the name of the source package. The ecosystem string might optionally have a :<RELEASE> suffix to scope the package to a particular Debian release. <RELEASE> is a numeric version specified in the [Debian distro-info-data](https://debian.pages.debian.net/distro-info-data/debian.csv). For example, the ecosystem string “Debian:7” refers to the Debian 7 (wheezy) release.

Would something similar make sense for Kubernetes? 

"Kubernetes" refers to upstream
"Kubernetes:<distribution>" refers to a specific distribution?

knqyf263: It sounds reasonable. Should we define a list of `<distribution>` here? It seems difficult to list all the K8s distributions, so can we simply show an example, like `“Kubernetes:<distribution>”, e.g. “Kubernetes:GKE”` or something like that?

oliverchang: I think we should define them, at least for the ones where we know there is a vulnerability database published. OSV's goal is to make vulnerability advisories machine readable and consistent.

oliverchang: @knqyf263 would you be able to help with defining a few distributions here and modifying this PR? We can start with the popular ones and add more later as required if we get feeds for them.

knqyf263: I found only two distributions publishing advisories. Do you know any other ones?

- [GKE](https://cloud.google.com/kubernetes-engine/security-bulletins)
- [Rancher](https://ranchermanager.docs.rancher.com/reference-guides/rancher-security/security-advisories-and-cves)

AKS seems to fix vulnerabilities [on GitHub](https://github.com/Azure/AKS/issues?q=is%3Aissue%20CVE%20label%3Asecurity), but it doesn't look like security advisories.

knqyf263: As far as I know, there are no machine-readable feeds. That's why we started https://github.com/kubernetes-sigs/cve-feed-osv with its definition.

>If these don't exist, maybe we should just start with plain "Kubernetes" to refer to upstream for now, and clarify it in the schema text?

I agree.

>Would having the name be the Go module name (i.e. k8s.io/apiserver) be the best way to encode upstream Kubernetes vulnerabilities (with the caveat that version numbers are different from the Go module as you pointed out https://github.com/kubernetes-sigs/cve-feed-osv/issues/9#issuecomment-2506104722)?

Another problem with Go module names is that the component not written in Go will not fit. We may want to take a list from [the official document](https://kubernetes.io/docs/concepts/overview/components/).

- kube-apiserver
- etcd
- kube-scheduler
- kube-controller-manager
- cloud-controller-manager
- kubelet
- kube-proxy

However, it's inconsistent with module names, `kube-apiserver` vs `k8s.io/apiserver`, and `kube-scheduler` vs `k8s.io/kube-scheduler`. I'm not sure which is better.

oliverchang: Hmm, [the official document](https://kubernetes.io/docs/concepts/overview/components/) doesn't seem to list all the possible components from past Kubernetes CVEs.

e.g. 

- ingress-nginx: https://github.com/kubernetes/kubernetes/issues/126744
- azure-file-csi-driver: https://github.com/kubernetes/kubernetes/issues/124759
- kubectl: https://github.com/kubernetes/kubernetes/issues/87773

I couldn't find any other document that comprehensively lists this either. It seems better to me to use the Go module names since they already exist and are consistent. 

e.g. k8s.io/ingress-nginx, sigs.k8s.io/azurefile-csi-driver, k8s.io/kubectl

WDYT?

knqyf263: >It seems better to me to use the Go module names since they already exist and are consistent.

Yes, that's the conclusion we also made when we started [cve-feed-osv](https://github.com/kubernetes-sigs/cve-feed-osv).

>Another problem with Go module names is that the component not written in Go will not fit.

What do you think about this problem? We also considered it but didn't develop a better solution than Go module names. Then, we decided to go with module names. I think it's a good time to discuss it with the OSV team.

oliverchang: > > Another problem with Go module names is that the component not written in Go will not fit.
> 
> What do you think about this problem? We also considered it but didn't develop a better solution than Go module names. Then, we decided to go with module names. I think it's a good time to discuss it with the OSV team.

Are there examples of such components today? 

We can probably also define this in the schema:

e.g. "If the component is a Go module, then the package name is the Go module name.".

knqyf263: As far as I know, there are non-Go projects under [the Kubernetes organization](https://github.com/kubernetes). However, I just went through [the Kubernetes SIGs organization](https://github.com/kubernetes-sigs) and found some non-Go projects.

- https://github.com/kubernetes-sigs/blixt
- https://github.com/kubernetes-sigs/nat64
- https://github.com/kubernetes-sigs/security-profiles-operator

I didn't find [a document](https://github.com/kubernetes/committee-security-response/blob/main/cna-handbook.md) on which organizations the K8s maintainers are responsible, but given that a vulnerability in [kubernetes-sigs/azure-file-csi-driver is announced](https://github.com/kubernetes/kubernetes/issues/124759), I presume SIGs projects can issue advisories.

oliverchang: Hmm, then it seems then the common denominator is just a git repo? i.e. 

```
{
  "package": {
    "ecosystem": "Kubernetes",
    "name": "https://github.com/kubernetes-sigs/blixt"
}

{
  "package": {
    "ecosystem": "Kubernetes",
    "name": "https://github.com/kubernetes/apiserver"
}
```

knqyf263: Their primary languages are not Go, but they have `go.mod` with Go module names.

- https://github.com/kubernetes-sigs/blixt/blob/0858e9284dd1c6c9a46f81b24f9f7d8664c51290/go.mod#L1
- https://github.com/kubernetes-sigs/security-profiles-operator/blob/b6741583f397f1349b5d4b4c30aa680eee2f2065/go.mod#L1
- https://github.com/kubernetes-sigs/nat64/blob/3a0e43b7eacfb60d84ed507b098dcc1501b853a7/go.mod#L1

And there are 184 repositories under `kubernetes-sigs`, but I found only three non-Go ones. Therefore, it's a very rare case.

"If the component has a Go module name, then the package name is the Go module name. Otherwise, use a git repo."

oliverchang: > Their primary languages are not Go, but they have go.mod with Go module names.

Interesting..

> "If the component has a Go module name, then the package name is the Go module name. Otherwise, use a git repo."

In the case of these 3 exceptions, since a go.mod exists, it sounds like should also use the Go module name under this definition? 

Since there is no case where there is a no go.mod today, I'd suggest just keeping this simple for now. We can always add additional cases later to the schema when we need them.

knqyf263: >In the case of these 3 exceptions, since a go.mod exists, it sounds like should also use the Go module name under this definition?

Exactly. That's what I wanted to express in the description.

>Since there is no case where there is a no go.mod today, I'd suggest just keeping this simple for now. We can always add additional cases later to the schema when we need them.

It sounds reasonable to me. Updated in https://github.com/ossf/osv-schema/pull/319/commits/01721af092a50925fd89c31e13294f58d4b40a95.

oliverchang (commented):
thanks! just a suggested rewording to make this a bit simpler

docs/schema.md
```diff
@@ -706,6 +717,7 @@ The defined ecosystems are:
 | `Go` | The Go ecosystem; the `name` field is a Go module path. |
 | `Hackage` | The Haskell package ecosystem. The `name` field is a Haskell package name as published on Hackage. |
 | `Hex` | The package manager for the Erlang ecosystem; the `name` is a Hex package name. |
+| `Kubernetes` | The Kubernetes ecosystem; If the component is a Go module, then the `name` filed is the Go module name. |
```
oliverchang, line 720:
```suggestion
| `Kubernetes` | The Kubernetes ecosystem; name is the Go module name associated with the relevant Kubernetes component (e.g. `k8s.io/apiserver`) |
```

oliverchang (commented):
(not sure where my suggestion went. reapplying)

docs/schema.md
knqyf263: Thanks. Applied.

oliverchang (approved):

PR #322, opened 2024-12-17
feat(linter): expand package checks across multiple ecosystems

hogo6002:
This PR enhances the linter by adding package existence checks for a variety of ecosystems:
- crates.io
- npm
- NuGet
- RubyGems
- Packagist
- Pub
- Hackage
- Maven

To accommodate these additions, the `ecosystems.go` file has been refactored, with code related to package existence checks moved to `package_check.go` and version checks moved to `version_check.go`. 

A current limitation is that malicious package record validation may fail, as these packages are often removed from package registries and are no longer queryable via API.

### DIFF
--- a/tools/osv-linter/internal/pkgchecker/ecosystems.go
+++ b/tools/osv-linter/internal/pkgchecker/ecosystems.go
@@ -66,7 +66,7 @@ func ExistsInEcosystem(pkg string, ecosystem string) bool {
 	case "Linux":
 		return true
 	case "Maven":
-		return true
+		return existsInMaven(pkg)
 	case "npm":
 		return existsInNpm(pkg)
 	case "NuGet":
--- a/tools/osv-linter/internal/pkgchecker/ecosystems.go
+++ b/tools/osv-linter/internal/pkgchecker/ecosystems.go
@@ -2,26 +2,34 @@ package pkgchecker
 
 import (
 	"fmt"
-	"io"
-	"net/http"
-	"regexp"
-	"slices"
-	"strings"
-
-	"golang.org/x/mod/module"
-	"golang.org/x/mod/semver"
-
-	"github.com/ossf/osv-schema/linter/internal/faulttolerant"
-
-	pep440 "github.com/aquasecurity/go-pep440-version"
-
-	"github.com/tidwall/gjson"
 )
 
 // Ecosystem support is a work in progress.
 var SupportedEcosystems = []string{
 	"Go",
 	"PyPI",
+	"crates.io",
+	"npm",
+	"NuGet",
+	"RubyGems",
+	"Packagist",
+	"Pub",
+	"Hackage",
+	"Maven",
+}
+
+// EcosystemBaseURLs maps ecosystems to their base API URLs.
+var EcosystemBaseURLs = map[string]string{
+	"Go":        "https://proxy.golang.org",
+	"PyPI":      "https://pypi.org/pypi",
+	"crates.io": "https://crates.io/api/v1/crates",
+	"npm":       "https://registry.npmjs.org",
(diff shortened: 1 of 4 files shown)


tools/osv-linter/internal/pkgchecker/ecosystems.go
```diff
@@ -2,26 +2,34 @@ package pkgchecker
...
+	"Go":        "https://proxy.golang.org",
+	"PyPI":      "https://pypi.org/pypi",
+	"crates.io": "https://crates.io/api/v1/crates",
+	"npm":       "https://registry.npmjs.org",
+	"NuGet":     "https://api.nuget.org/v3-flatcontainer",
+	"RubyGems":  "https://rubygems.org/api/v1/gems",
+	"Packagist": "https://repo.packagist.org/p2",
+	"Pub":       "https://pub.dev/api/packages",
+	"Hackage":   "https://hackage.haskell.org/package",
+	"Maven":     "https://search.maven.org/solrsearch/select",
```
oliverchang, line 32: early initial comment: would it make sense to use deps.dev where it's available for an ecosystem? 

some of these have rate limits etc that might be problematic.

hogo6002: Yes, I thought about using deps.dev, but I was concerned about the comprehensiveness of its data. So originally I planned to use deps.dev or another source as a secondary check when the primary source failed with 500 errors. But yeah, rate limits are quite important. I will update code to use deps.dev as the primary check source and set the ecosystem-specific registries as secondary ones to bypass the rate limit issue.

tools/osv-linter/internal/pkgchecker/ecosystems.go
```diff
@@ -2,26 +2,34 @@ package pkgchecker
...
+	"NuGet",
+	"RubyGems",
+	"Packagist",
+	"Pub",
+	"Hackage",
+	"Maven",
+}
+
+// EcosystemBaseURLs maps ecosystems to their base API URLs.
+var EcosystemBaseURLs = map[string]string{
```
andrewpollock, line 22: There's an opportunity to DRY this out a bit, and we can use the keys of `EcosystemBaseURLs` to infer what is now duplicated into `SupportedEcosystems`.

tools/osv-linter/internal/pkgchecker/package_check.go
```diff
@@ -0,0 +1,153 @@
...
+	artifact_id := strings.Split(pkg, ":")[1]
+
+	ecosystem := "Maven"
+	packageInstanceURL := fmt.Sprintf("%s/?q=g:%s%%20AND%%20a:%s", EcosystemBaseURLs[ecosystem], group_id, artifact_id)
+
+	if isPackageInDepsDev(ecosystem, pkg) {
+		return true
+	}
+
+	// Needs to use GET instead of HEAD for Maven
```
andrewpollock, line 130: Please also add why to this comment

```suggestion
	// Needs to use GET instead of HEAD for Maven.
```

tools/osv-linter/internal/pkgchecker/package_check.go
andrewpollock, line 1: There's a lot of (understandably valid) inconsistency to the implementation of the functions in this package. To assist with contributions, please add a package docstring that explains the goals of functions in this package and provides a canonical example that contributors can crib from, to avoid getting cargo-culted contributions from a non-canonical function implementation.

andrewpollock (approved):
LGTM, thanks!

tools/osv-linter/internal/pkgchecker/ecosystems.go
hogo6002: oops, forgot about this PR.  I'll get to those comments in a follow-up one.

Issue #325, opened 2025-01-01
Link directly to CVSS vector string documentation


mjpitz:
The CVSS string format is extremely confusing for newcomers. Currently, the documentation for the severity score value all link to roughly the same location (https://www.first.org/cvss/) which doesn't provide any information about the vector string format whatsoever...

I think it would be far more valuable to actually link to the respective vector string documentation for each format.

- CVSS_V2 - https://www.first.org/cvss/v2/guide#Metric-Groups
- CVSS_V3 - https://www.first.org/cvss/v3.0/specification-document#Vector-String
- CVSS_V4 - https://www.first.org/cvss/v4.0/specification-document#Vector-String

mjpitz:
If folks think this is worth doing, I'm open to submitting a PR for y'all

oliverchang:
Thank you for the suggestion! This seems very reasonable to me and would be a great way to make this clearer to readers. Can you please open a PR?

PR #326, opened 2025-01-08
docs: point to vector string documentation

mjpitz:
Resolves https://github.com/ossf/osv-schema/issues/325


oliverchang (approved):
thank you!

PR #327, opened 2025-01-17
Update JSON schema to reflect top-level/package severities mutual exclusivity

litios:
`affected[].severity` mentions:

 'If any package level severity fields are set,
 the top level severity must not be set.'

This change updates the JSON schema under validation to reflect this.

Changes have been tested with both tools mentioned under validation/README.md

### DIFF
--- a/validation/schema.json
+++ b/validation/schema.json
@@ -295,6 +295,26 @@
     "id",
     "modified"
   ],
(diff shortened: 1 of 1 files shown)


oliverchang (approved):
thanks!

PR #333, opened 2025-02-04
feat: add go struct definitions of OSV record

michaelkedar:
Moving the struct definitions of the OSV record from [osv-scanner](https://github.com/google/osv-scanner/blob/b870d62b9987f27ab0212a9fd0eeef8086554699/pkg/models/vulnerability.go) to this repository, since it makes more sense to live in the `osvschema` module.

I've also added the new `upstream` field to the struct. It's probably worth having a discussion on how to make sure these structs remain in-sync with any future changes to the schema.

### DIFF
--- a/bindings/go/osvschema/vulnerability.go
+++ b/bindings/go/osvschema/vulnerability.go
@@ -10,9 +10,9 @@ import (
 //
 // See: https://ossf.github.io/osv-schema/#affectedpackage-field
 type Package struct {
-	Ecosystem Ecosystem `json:"ecosystem"      yaml:"ecosystem"`
-	Name      string    `json:"name"           yaml:"name"`
-	Purl      string    `json:"purl,omitempty" yaml:"purl,omitempty"`
+	Ecosystem string `json:"ecosystem"      yaml:"ecosystem"`
+	Name      string `json:"name"           yaml:"name"`
+	Purl      string `json:"purl,omitempty" yaml:"purl,omitempty"`
 }
 
 // Event describes a single version that either:
--- a/bindings/go/osvschema/vulnerability.go
+++ b/bindings/go/osvschema/vulnerability.go
@@ -65,27 +65,6 @@ type Affected struct {
 	EcosystemSpecific map[string]interface{} `json:"ecosystem_specific,omitempty" yaml:"ecosystem_specific,omitempty"`
 }
 
-// MarshalJSON implements the json.Marshaler interface.
-//
-// This method ensures Package is only present if it is not equal to the zero value.
-// This is achieved by embedding the Affected struct with a pointer to Package used
-// to populate the "package" key in the JSON object.
-func (a Affected) MarshalJSON() ([]byte, error) {
-	type rawAffected Affected // alias Affected to avoid recursion during Marshal
-	type wrapper struct {
-		Package *Package `json:"package,omitempty"`
-		rawAffected
-	}
-	raw := wrapper{rawAffected: rawAffected(a)}
-	if a.Package == (Package{}) {
-		raw.Package = nil
-	} else {
-		raw.Package = &(a.Package)
-	}
-
-	return json.Marshal(raw)
-}
-
 // Reference links to additional information, advisories, issue tracker entries,
 // and so on about the vulnerability itself.
 //
@@ -125,6 +104,27 @@ type Vulnerability struct {
 	References       []Reference            `json:"references,omitempty"        yaml:"references,omitempty"`
 	Credits          []Credit               `json:"credits,omitempty"           yaml:"credits,omitempty"`
 	DatabaseSpecific map[string]interface{} `json:"database_specific,omitempty" yaml:"database_specific,omitempty"`
+}
+
+// MarshalJSON implements the json.Marshaler interface.
+//
+// This method ensures Package is only present if it is not equal to the zero value.
+// This is achieved by embedding the Affected struct with a pointer to Package used
+// to populate the "package" key in the JSON object.
+func (a Affected) MarshalJSON() ([]byte, error) {
+	type rawAffected Affected // alias Affected to avoid recursion during Marshal
+	type wrapper struct {
+		Package *Package `json:"package,omitempty"`
+		rawAffected
+	}
+	raw := wrapper{rawAffected: rawAffected(a)}
+	if a.Package == (Package{}) {
+		raw.Package = nil
+	} else {
+		raw.Package = &(a.Package)
+	}
+
+	return json.Marshal(raw)
 }
 
 // MarshalJSON implements the json.Marshaler interface.
--- a/bindings/go/osvschema/vulnerability.go
+++ b/bindings/go/osvschema/vulnerability.go
@@ -5,7 +5,7 @@ import (
 	"time"
 )
 
-// Package identifies the affected code library or command provided by the
+// Package describes the affected code library or command provided by the
 // package.
 //
 // See: https://ossf.github.io/osv-schema/#affectedpackage-field
@@ -43,7 +43,7 @@ type Range struct {
 	DatabaseSpecific map[string]interface{} `json:"database_specific,omitempty" yaml:"database_specific,omitempty"`
 }
 
-// Severity is used to describe the severity of a vulnerability for an affected
+// Severity describes the severity of a vulnerability for an affected
 // package using one or more quantitative scoring methods.
 //
 // See: https://ossf.github.io/osv-schema/#severity-field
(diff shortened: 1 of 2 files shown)


bindings/go/osvschema/vulnerability.go
```diff
@@ -0,0 +1,170 @@
...
+	"encoding/json"
+	"time"
+)
+
+// Package identifies the affected code library or command provided by the
+// package.
+//
+// See: https://ossf.github.io/osv-schema/#affectedpackage-field
+type Package struct {
+	Ecosystem Ecosystem `json:"ecosystem"      yaml:"ecosystem"`
```
michaelkedar, line 13: I'm not sure if using the `Ecosystem` type is technically correct here. The constants are defined without the `:<RELEASE>` suffixes that ecosystems use
(e.g. `EcosystemAlpine Ecosystem = "Alpine"`, but this would hold `Alpine:v3.16`)

another-rex: Ecosystem should just be string here I believe. On osv-scanner's side we have the Parsed type, but that's not available in osv-schema (since it requires some opinionated decisions.)

michaelkedar: okay, I've made it a string

bindings/go/osvschema/vulnerability.go
```diff
@@ -0,0 +1,170 @@
...
+	Severity         []Severity             `json:"severity,omitempty"          yaml:"severity,omitempty"`
+	Affected         []Affected             `json:"affected,omitempty"          yaml:"affected,omitempty"`
+	References       []Reference            `json:"references,omitempty"        yaml:"references,omitempty"`
+	Credits          []Credit               `json:"credits,omitempty"           yaml:"credits,omitempty"`
+	DatabaseSpecific map[string]interface{} `json:"database_specific,omitempty" yaml:"database_specific,omitempty"`
+}
+
+// MarshalJSON implements the json.Marshaler interface.
+//
+// This method ensures times all times are formatted correctly according to the schema.
```
andrewpollock, line 132: Correct comment grammar

```suggestion
// This method ensures all times are formatted correctly according to the schema.
```

bindings/go/osvschema/vulnerability.go
```diff
@@ -0,0 +1,170 @@
+package osvschema
+
+import (
+	"encoding/json"
+	"time"
+)
+
+// Package identifies the affected code library or command provided by the
```
andrewpollock, line 8: Standardize on the verb for consistency

```suggestion
// Package describes the affected code library or command provided by the
```

bindings/go/osvschema/vulnerability.go
```diff
@@ -0,0 +1,170 @@
...
+//
+// See: https://ossf.github.io/osv-schema/#affectedranges-field
+type Range struct {
+	Type             RangeType              `json:"type"                        yaml:"type"`
+	Events           []Event                `json:"events"                      yaml:"events"`
+	Repo             string                 `json:"repo,omitempty"              yaml:"repo,omitempty"`
+	DatabaseSpecific map[string]interface{} `json:"database_specific,omitempty" yaml:"database_specific,omitempty"`
+}
+
+// Severity is used to describe the severity of a vulnerability for an affected
```
andrewpollock, line 46: Standardize for consistency.

```suggestion
// Severity describes the severity of a vulnerability for an affected
```

bindings/go/osvschema/vulnerability.go
```diff
@@ -0,0 +1,170 @@
...
+	Severity          []Severity             `json:"severity,omitempty"           yaml:"severity,omitempty"`
+	Ranges            []Range                `json:"ranges,omitempty"             yaml:"ranges,omitempty"`
+	Versions          []string               `json:"versions,omitempty"           yaml:"versions,omitempty"`
+	DatabaseSpecific  map[string]interface{} `json:"database_specific,omitempty"  yaml:"database_specific,omitempty"`
+	EcosystemSpecific map[string]interface{} `json:"ecosystem_specific,omitempty" yaml:"ecosystem_specific,omitempty"`
+}
+
+// MarshalJSON implements the json.Marshaler interface.
+//
+// This method ensures Package is only present if it is not equal to the zero value.
```
andrewpollock, line 70: Is it possible to also implicitly meet or explicitly enforce the need for an `introduced` event here? https://ossf.github.io/osv-schema/#requirements

I'm leaning towards implicitly adding a zero value in the absence of an explicit one over failing when one's absent, for user-friendliness, even though I'm normally very anti implicit behaviours. I'm not seeing any immediate downside from this approach...

bindings/go/osvschema/vulnerability.go
```diff
@@ -0,0 +1,170 @@
...
+	if !v.Withdrawn.IsZero() {
+		raw.Withdrawn = v.Withdrawn.UTC().Format(time.RFC3339)
+	}
+
+	return json.Marshal(raw)
+}
+
+// MarshalYAML implements the yaml.Marshaler interface.
+//
+// This method ensures times all times are formatted correctly.
```
andrewpollock, line 155: Make consistent with `MarshalJSON()`

```suggestion
// This method ensures times all times are formatted correctly according to the schema.
```

bindings/go/osvschema/vulnerability.go
```diff
@@ -0,0 +1,170 @@
...
+	}
+
+	return json.Marshal(raw)
+}
+
+// Reference links to additional information, advisories, issue tracker entries,
+// and so on about the vulnerability itself.
+//
+// See: https://ossf.github.io/osv-schema/#references-field
+type Reference struct {
```
andrewpollock, line 93: Could we consolidate all of the types before the funcs, for readability?

bindings/go/osvschema/vulnerability.go
```diff
@@ -0,0 +1,170 @@
...
+	Type ReferenceType `json:"type" yaml:"type"`
+	URL  string        `json:"url"  yaml:"url"`
+}
+
+// Credit gives credit for the discovery, confirmation, patch, or other events
+// in the life cycle of a vulnerability.
+//
+// See: https://ossf.github.io/osv-schema/#credits-fields
+type Credit struct {
```
andrewpollock, line 102: Could we consolidate all of the types before the funcs?

bindings/go/osvschema/vulnerability.go
```diff
@@ -0,0 +1,170 @@
...
+// Reference links to additional information, advisories, issue tracker entries,
+// and so on about the vulnerability itself.
+//
+// See: https://ossf.github.io/osv-schema/#references-field
+type Reference struct {
+	Type ReferenceType `json:"type" yaml:"type"`
+	URL  string        `json:"url"  yaml:"url"`
+}
+
+// Credit gives credit for the discovery, confirmation, patch, or other events
```
andrewpollock, line 98: Standardize for consistency.

```suggestion
// Credit describes who to give credit to for the discovery, confirmation, patch, or other events
```

andrewpollock (commented):
I personally think this is the right place to consolidate such code, so thanks for creating this PR. Some readability feedback.

bindings/go/osvschema/vulnerability.go
michaelkedar: I think it's possible, but there's not really any other correctness enforcing in these structs. e.g. there's nothing stopping a single event from containing both `introduced` and `fixed` fields.

I think if someone constructs an invalid struct, it's okay if it produces an invalid record.

michaelkedar:
> LGTM. I'm guessing the reason we don't need unmarshal functions is because the default one should just work?

Yeah, the custom marshal methods are just for omitting empty struct fields and making sure timestamps are in UTC - there's no customized encoding going on.

bindings/go/osvschema/vulnerability.go
andrewpollock: What I'm thinking of is essentially [Postel's law](https://lawsofux.com/postels-law/). I'm not going to die on the hill over it, but I think it's a sign of a higher-quality piece of software, and you don't need to boil the ocean on achieving it comprehensively here and now, but this could be a good starting point... Your call.

andrewpollock (approved):

bindings/go/osvschema/vulnerability.go
michaelkedar: I had a bit more of a think about this, and I think I'd prefer not to do this (at least, not *just* this in this PR).

Generally, I'm against inserting `{"introduced": "0"}` because it's not what the implied behaviour is (empty ranges / ranges with only `fixed` events would match nothing). It'd cause Marshalling/Unmarshalling a struct to represent different ranges.

I'd rather return an error, but it'd be a bit misleading if it only errors in this one case - if it *can* error due to being invalid and it *doesn't*, then I'd assume it produced a valid OSV (even though there are many other ways for it to be invalid).

If you care about producing a correctly-formatted OSV record, you'd still need to check it against the schema. Adding just this one check doesn't really provide much value.

michaelkedar:
I can't merge this PR myself - I'm happy for this to be merged if you are.

PR #335, opened 2025-02-04
Add V8 prefix.

vsutedjo:
### Add V8 prefix to OSV-schema.
As per internal discussions, we are adding the "V8" prefix for policy-based synthetic vulnerability feeds to OSV schema.
The Chromium team has a [policy](https://chromium.googlesource.com/chromium/src/+/refs/heads/main/docs/security/updates.md) that states that direct embedders of Chromium packages such as V8 are responsible for using an up-to-date version of V8 packages. The definition of that is that the V8 package is at most one week old.

Since these types of packages do not have their own identifiers and therefore no advisory could be created, enforcement of such policies was not possible. We now created a synthetic advisory generator that takes such time-based policies and converts them into regularily updated advisories that OSV can read:
https://github.com/google/chromium-policy-vulnfeed/blob/main/advisories/V8-advisory.json


andrewpollock:
Are you able to attend to https://github.com/ossf/osv-schema/runs/36678661488 ?

docs/schema.md
```diff
@@ -454,6 +454,17 @@ The defined database prefixes and their "home" databases are:
...
       </td>
     </tr>
+    <tr>
+      <td><code>V8</code></td>
+      <td><a href="https://chromium.googlesource.com/chromium/src/+/refs/heads/main/docs/security/updates.md">V8/Chromium Time-Based Policy</a></td>
+      <td>
+        <ul>
+          <li>How to contribute: TBD</li>
+          <li>Source URL: TBD</li>
+          <li>OSV Formatted URL: <code>N/A</code></li>
```
oliverchang, line 464: would it make sense to link to https://github.com/google/chromium-policy-vulnfeed/blob/main/advisories/V8-advisory.json for both Source and OSV Formatted URL?

oliverchang (commented):
thanks! +1 to Andrew's comment on the DCO check.

docs/schema.md
vsutedjo: Done.

vsutedjo:
I think i borked my commit history with trying to add the sign-offs just now, I just re-started the process here. https://github.com/ossf/osv-schema/pull/339

Will close this one.

PR #345, opened 2025-03-18
tools: Update redhat csaf2osv conversion to support upstream field + tests

jess-lowe:
CVE upstream has moved from `related` to `upstream` to support https://github.com/ossf/osv-schema/pull/312 schema change. 

The `GO` vuln alias (for the CVE entry) has been left in related, so as not to affect the computation on the OSV.dev hierarchy display side. 

Tests updated to the latest schema version as well.


jasinner:
Looks good, thank you!

andrewpollock (approved):

PR #352, opened 2025-05-20
feat(osv-linter): add checks for upstream and aliases fields

hogo6002:
This PR adds some new validation checks for upstream, aliases, and related fields to help linux distro data validation

It also made several changes to streamline linter integration into the OSV data import pipeline:

- Error Code Refactoring: Error codes have been updated (e.g. from R0001 to RNG:001). This provides better clarity and scalability as we add more checks (e.g. distinguishing Range, Record, or Relation errors). The colon delimiter also simplifies parsing for users.

- fatal Check Collection: A fatal check collection is used to categorize critical validation failures. 

- JSON Output: Added a JSON output format to simplify programmatic parsing of validation results.

- added MinimOS ecosystem

### DIFF
--- a/tools/osv-linter/internal/checks/record.go
+++ b/tools/osv-linter/internal/checks/record.go
@@ -58,39 +58,23 @@ func UpstreamCheck(json *gjson.Result, config *Config) (findings []CheckError) {
 	}
 
 	aliases := json.Get("aliases")
-	if !aliases.Exists() {
-		return
-	}
-
-	hasAliases := false
-	if aliases.IsArray() {
+	if aliases.Exists() && aliases.IsArray() {
 		for _, bug := range aliases.Array() {
 			if _, exists := unique[bug.String()]; exists {
-				hasAliases = true
+				findings = append(findings, CheckError{Message: "Invalid Upstream: upstream should not contain aliases"})
+				break
 			}
 		}
-	}
-
-	if hasAliases {
-		findings = append(findings, CheckError{Message: "Invalid Upstream: upstream should not contain aliases"})
 	}
 
 	related := json.Get("related")
-	if !related.Exists() {
-		return
-	}
-
-	hasRelated := false
-	if related.IsArray() {
+	if related.Exists() && related.IsArray() {
 		for _, bug := range related.Array() {
 			if _, exists := unique[bug.String()]; exists {
-				hasRelated = true
+				findings = append(findings, CheckError{Message: "Invalid Upstream: upstream should not contain related IDs"})
+				break
 			}
 		}
-	}
-
-	if hasRelated {
-		findings = append(findings, CheckError{Message: "Invalid Upstream: upstream should not contain related IDs"})
 	}
 
 	return findings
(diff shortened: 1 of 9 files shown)


tools/osv-linter/internal/checks/record.go
```diff
@@ -19,3 +40,114 @@ func RecordHasAffected(json *gjson.Result, config *Config) (findings []CheckErro
...
+		for _, bug := range aliases.Array() {
+			if _, exists := unique[bug.String()]; exists {
+				hasAliases = true
+			}
+		}
+	}
+
+	if hasAliases {
+		findings = append(findings, CheckError{Message: "Invalid Upstream: upstream should not contain aliases"})
+	}
```
michaelkedar, line 76: nit: `hasAliases` is not really a good name - this is more `aliasInUpstream`.
But also, you can skip the `hasAliases` variable and just do
```go
if _, exists := unique[bug.String()]; exists {
	findings = append(findings, CheckError{Message: "Invalid Upstream: upstream should not contain aliases"})
	break
}
```

same with `hasRelated` below

tools/osv-linter/internal/checks/record.go
```diff
@@ -19,3 +40,114 @@ func RecordHasAffected(json *gjson.Result, config *Config) (findings []CheckErro
...
+			bugIDs = append(bugIDs, bugResult.String())
+		}
+	} else {
+		return false, nil
+	}
+
+	seen := make(map[string]bool)
+	for _, bugID := range bugIDs {
+		seen[bugID] = true
+	}
```
michaelkedar, line 150: nit: it's mildly more efficient to use `map[string]struct{}` for this

tools/osv-linter/internal/checks/record.go
```diff
@@ -19,3 +40,114 @@ func RecordHasAffected(json *gjson.Result, config *Config) (findings []CheckErro
...
+	hasRelated := false
+	if related.IsArray() {
+		for _, bug := range related.Array() {
+			if _, exists := unique[bug.String()]; exists {
+				hasRelated = true
+			}
+		}
+	}
+
+	if hasRelated {
```
jess-lowe, line 92: I'm not sure that the linter needs to fatal on related and upstream having overlap. It's not ideal, but it shouldn't fatal.

jess-lowe: Maybe move this check into the RelatedCheck so it isn't fatal?

hogo6002: Ideally, we should add a warning type to it, but this requires a bit of design and discussion. So for this week, I will be focusing on getting the pipeline to work first. 

I understand that it doesn't matter what to put in the related field. But I still like to keep the check here. As upstream is now being used by Linux distro maintainers for the first time, I'd like to let people understand the usage of upstream. It would be best for new data sources to convert correctly in the first place, rather than correcting them in the future. 
But if this really bothers people, we can move it into another category in the future. So I would probably keep it there unless we really see a case where a bug id has to exist in both the related and upstream fields at the same time.

oliverchang (approved):

PR #354, opened 2025-05-22
fix(linter): address linter issues on withdrawn records and range version and purl

hogo6002:
Fixes linter issues by:
- Skipping affected field checks for withdrawn records.
- Allowing records to have matching last_affected and introduced versions.
- Adding a mock version to ecosystem PURL strings when required

### DIFF
--- a/tools/osv-linter/internal/checks/packages.go
+++ b/tools/osv-linter/internal/checks/packages.go
@@ -162,7 +162,12 @@ func PackagePurlValid(json *gjson.Result, config *Config) (findings []CheckError
 
 		_, err := packageurl.FromString(purl.String())
 		if err != nil {
-			findings = append(findings, CheckError{Message: fmt.Sprintf("Invalid Purl %q: %#v", purl.String(), err)})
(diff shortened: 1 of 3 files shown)


hogo6002:
cc@andrewpollock 

We're doing a fix week, and I'm trying to integrate OSV-linter into the importer pipeline and adding a cron job to regularly run the linter on historical records. I made some changes on OSV-linter, as this PR and https://github.com/ossf/osv-schema/pull/352, to support this integration. Now, the linter has found 48k records with issues in the test instance, and these can be queried via http://api.test.osv.dev/v1experimental/importfindings/ API. Let me know if you have any feedback or suggestions on the changes I made.

another-rex (approved):
Please update the title to cover all 3 fixes, or just say general fixes, right now I think it might be a bit confusing when glancing over the changes in git history for example.

another-rex (approved):
Please update the title to cover all 3 fixes, or just say general fixes, right now I think it might be a bit confusing when glancing over the changes in git history for example.

PR #358, opened 2025-06-12
add openEuler ecosystem

kirigiricloud:
Add support for openEuler ecosystem


ecosystems.json
```diff
@@ -21,6 +21,7 @@
   "MinimOS": "The MinimOS package ecosystem; the `name` is the name of the package.",
   "npm": "The NPM ecosystem; the `name` field is an NPM package name.",
   "NuGet": "The NuGet package ecosystem. The `name` field is a NuGet package name.",
+  "openEuler": "The openEuler ecosystem; The `name` field is the name of the source RPM. The ecosystem string has a `<RELEASE>` suffix, specifying a particular openEuler LTS Release.`<RELEASE>` is numeric (YY.MM) version maintained in our [archive list](https://www.openeuler.org/en/download/?archive=true). Here, `LTS` stands for long term support and `SP` stands for service pack which offers extensions and enhancements of the major LTS version. Note innovation versions (those without `LTS`) are out of our security advisories' scope. There is an `ecosystem_specific` array covers all updated packages, including src rpm and binaries of different architectures.",
```
oliverchang, line 24:
```suggestion
  "openEuler": "The openEuler ecosystem; The `name` field is the name of the source RPM. The ecosystem string has a `<RELEASE>` suffix, specifying a particular openEuler LTS Release.`<RELEASE>` is numeric (YY.MM) version maintained in our [archive list](https://www.openeuler.org/en/download/?archive=true). Here, `LTS` stands for long term support and `SP` stands for service pack which offers extensions and enhancements of the major LTS version. Note innovation versions (those without `LTS`) are out of our security advisories' scope. The `ecosystem_specific` field contains all updated packages, including src rpm and binaries of different architectures.",
```

oliverchang: Note that ecosystem_specific has to be an object, not an array. 

Do you also have a link to the definition of `ecosystem_specific` that we can link from here?

oliverchang (approved):
nice! just one minor comment

oliverchang (approved):

PR #365, opened 2025-07-14
feat: add json schema validation into linter

hogo6002:
Adds JSON schema validation to the linter.

details: https://github.com/google/osv.dev/issues/3666


tools/osv-linter/internal/checks/schema.json
```diff
@@ -0,0 +1,494 @@
+{
```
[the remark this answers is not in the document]
hogo6002: Updated. the only downside is that //go:embed can't directly embed symlink links. So it can only read the file at runtime

chrisbloom7 (commented):
Great addition. Thanks! (leaving for others to approve since I've been away from the project for a bit)

tools/osv-linter/internal/linter.go
```diff
@@ -40,6 +40,10 @@ func lint(content *Content, config *Config) (findings []checks.CheckError) {
 		log.Printf("%q: invalid JSON", content.filename)
 	}
 
+	if !checks.ValidateJSON(string(content.bytes), content.filename, config.verbose) {
+		findings = append(findings, checks.CheckError{Code: checks.CheckInvalidSchema.Code, Message: "Invalid JSON"})
```
another-rex, line 44: Should the message contain the schema that actually occured?

Though either way I think them message shouldn't be invalid JSON, since json is likely valid, just doesn't pass the schema.

tools/osv-linter/internal/linter.go
```diff
@@ -40,6 +40,10 @@ func lint(content *Content, config *Config) (findings []checks.CheckError) {
 		log.Printf("%q: invalid JSON", content.filename)
```
another-rex, line 40: Shouldn't this exit early with a Invalid JSON error?

tools/osv-linter/internal/checks/schema.go
```diff
@@ -0,0 +1,48 @@
...
+
+	schemaLoader := gojsonschema.NewReferenceLoader("file://" + schemaFilePath)
+
+	// Load the JSON data to be validated
+	documentLoader := gojsonschema.NewStringLoader(jsonData)
+
+	// Perform the validation
+	result, err := gojsonschema.Validate(schemaLoader, documentLoader)
+	if err != nil {
+		log.Fatalf("error during validation: %s", err)
```
another-rex, line 33: Make the panic error message clear that both the input JSON and the schema should be valid before entering this function.

tools/osv-linter/internal/linter.go
another-rex: Otherwise we'll panic in the ValidateJSON function.

another-rex (changes requested):
Where is the linter being ran, and will it get access to the schema (in the right file path) when it is executed? I have a feeling we don't copy in the source code (the symlink), nor the target of the symlink, when this is executed. 

A better solution might be to create a `go generate` file to copy in the schema.json file (and add the copied file to the .gitignore). Then use go:embed to embed the json data.

another-rex (approved):
LGTM

PR #374, opened 2025-07-29
test: ensure that schema has been (re)generated

G-Rath:
This should help ensure the two files are kept in sync - while this doesn't feel the best, it seemed slightly better than e.g. having a job that does a `diff` since we already run tests in a dedicated job


G-Rath:
This will fail until #375 is landed

another-rex (approved):

PR #382, opened 2025-07-30
refactor: use `osv-scalibr/semantic` for version comparing

G-Rath:
Branched off #381

---

The `semantic` package supports versions more ecosystems and is being used by related OSV tools so it makes sense to use it here ahead of introducing support for more ecosystems


another-rex (approved):
Thanks!

PR #389, opened 2025-08-04
feat: support checking if versions exist in Hackage

G-Rath:
This adds support for checking that Hackage packages exist in the repository

### DIFF
--- a/tools/osv-linter/internal/pkgchecker/version_check.go
+++ b/tools/osv-linter/internal/pkgchecker/version_check.go
@@ -216,6 +216,18 @@ func goVersionsExist(versions []string) error {
 	return nil
 }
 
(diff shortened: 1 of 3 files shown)


tools/osv-linter/internal/pkgchecker/version_check.go
```diff
@@ -129,6 +129,58 @@ func goVersionsExist(versions []string) error {
...
+				break
+			}
+		}
+		if versionFound {
+			continue
+		}
+		versionsMissing = append(versionsMissing, versionToCheckFor)
+	}
+	if len(versionsMissing) > 0 {
+		return &MissingVersionsError{Package: pkg, Ecosystem: "RubyGems", Missing: versionsMissing, Known: versionsInRepository}
```
another-rex, line 178: Same here

PR #391, opened 2025-08-04
feat: support checking if versions exist in crates.io

G-Rath:
This adds support for checking that crates.io packages exist in the repository

### DIFF
--- a/tools/osv-linter/internal/pkgchecker/version_check.go
+++ b/tools/osv-linter/internal/pkgchecker/version_check.go
@@ -44,7 +44,7 @@ func versionsExistInCrates(pkg string, versions []string) error {
 	versionsMissing := []string{}
 	for _, versionToCheckFor := range versions {
 		versionFound := false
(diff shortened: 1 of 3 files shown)


tools/osv-linter/internal/pkgchecker/version_check.go
```diff
@@ -15,6 +15,58 @@ import (
...
+	releases := gjson.GetBytes(respJSON, "versions")
+	releases.ForEach(func(key, value gjson.Result) bool {
+		versionsInRepository = append(versionsInRepository, value.Get("num").String())
+		return true // keep iterating.
+	})
+	// Determine which referenced versions are missing.
+	versionsMissing := []string{}
+	for _, versionToCheckFor := range versions {
+		versionFound := false
+		vc, err := semantic.Parse(versionToCheckFor, "RubyGems")
```
another-rex, line 47: This is RubyGems instead of cargo?

G-Rath: 😬

PR #407, opened 2025-08-19
feat: support checking versions exist in Pub

G-Rath:


G-Rath:
@another-rex I found one more to do 😅 though this is blocked until https://github.com/google/osv-scalibr/pull/1196 is landed

PR #417, opened 2025-09-09
feat: add `vulnerability.proto` and generated bindings

cuixq:
This PR adds `vulnerability.proto` to closely bind it with the schema, as well as adds the generated bindings for Go and Python so we don't need to manually maintain the bindings.

For Go bindings, this replaces the existing `vulnerability.go` and there will be another PR soon for `constants.go`.


another-rex (approved):
Nice!

PR #421, opened 2025-09-15
feat: add support for `ALPINE-` prefix

jess-lowe:
OSV.dev is looking to decouple Debian and Alpine CVE Trackers from its main CVE-record.

This PR adds the ALPINE- prefix for use for CVEs tracked by Alpine's security tracker to the schema.


docs/schema.md
```diff
@@ -179,6 +179,17 @@ The defined database prefixes and their "home" databases are:
         </ul>
       </td>
     </tr>
+    <tr>
+      <td><code>ALPINE</code></td>
+      <td><a href="https://secdb.alpinelinux.org/">Alpine</a></td>
```
another-rex, line 184: Add provided by osv.dev similar to debian.

PR #424, opened 2025-09-18
Adjust to the new RPM Module format in CSAF VEX

jasinner:
Red Hat changed the way RPMs shipped in modules are described in CSAF files. This change reflects the new format.


jasinner:
Here are the 2 CSAF files you mentioned regenerated with the updated code in this branch:

tools/redhat/redhat_osv/csaf.py
```diff
@@ -2,9 +2,59 @@
...
+        Note: The version is URL-decoded (e.g., %2B becomes +)
+
+    Raises:
+        ValueError: If the PURL format is invalid or not an RPM PURL
+    """
+    if not purl.startswith("pkg:rpm/"):
+        raise ValueError(f"Only RPM PURLs are supported, got: {purl}")
+
+    # Parse the PURL: pkg:rpm/namespace/name@version?qualifiers
+    parsed = urlparse(purl)
```
michaelkedar, line 30: Perhaps using [packageurl-python](https://github.com/package-url/packageurl-python) instead of urllib & manual splitting might be more reliable

tools/redhat/redhat_osv/csaf.py
```diff
@@ -30,42 +80,58 @@ def __post_init__(self, csaf_product_id: str, cpes: dict[str, str],
...
+            if self.product_version.count(":") >= 4:
+                # Find the 4th colon and take everything after it as the NEVRA
+                colon_positions = [
+                    i for i, char in enumerate(self.product_version)
+                    if char == ':'
+                ]
+                if len(colon_positions) >= 4:
+                    nevra_start = colon_positions[
+                        3] + 1  # Start after the 4th colon
+                    nevra_candidate = self.product_version[nevra_start:]
```
michaelkedar, line 103: nit: It seems like it'd be cleaner to use something like
```python
nevra_candidate = self.product_version.split(':', maxsplit=4)[4]
```

tools/redhat/setup.py
```diff
@@ -2,12 +2,9 @@
...
 """
 
-from setuptools import setup, find_packages
+from setuptools import setup
 
-REQUIRES = [
-    "jsonschema",
-    "requests"
-]
+REQUIRES = ["jsonschema", "requests"]
```
michaelkedar, line 7: I think you need to add `urllib3` (or `packageurl-python`) here?

michaelkedar (approved):
LGTM
A couple of code comments for your consideration (but I won't mind if you want to ignore)

I don't think I have permissions here - @another-rex could you PTAL

another-rex (commented):
LGTM, though +1 on the packageurl comment instead of manually parsing it.

tools/redhat/redhat_osv/csaf.py
jasinner: Ok, I'm using packageurl-python now.

tools/redhat/redhat_osv/csaf.py
jasinner: Agreed. I added code to that affect.

tools/redhat/setup.py
jasinner: Ok, updated that too, thanks.

PR #425, opened 2025-09-19
Update ghsa tool Pipfile to grab latest OSV so JSON keys are sorted.

calebbrown:
This is to help fix https://github.com/ossf/malicious-packages/issues/1002


calebbrown:
Appears to have already been done.

PR #438, opened 2025-10-14
add small clarification to affected versions

michaelkedar:
Recently, in OSV.dev (https://github.com/google/osv.dev/issues/4143) someone pointed out that the 'fixed' version of some of our records do not map directly to an upstream package version.

Update the schema to clarify that this may be the case.


another-rex:
> Also, I'd think it may be a good idea to limit this relaxation to SemVer ecosystems where we know exactly how to compare versions, even if they are fictitious.

I think currently it's ambiguous, so if we just specify semver, it's still ambiguous for non semver ecosystems (or imply that we are restricting the requirements on non semver ranges from the current state).

PR #450, opened 2025-11-14
Update VSCode description to allow for registries to be specified in the ecosystem

### LINKED ISSUE #449: Specifying registries for VSCode ecosystem
I am looking at the VSCode support recently added in #357 and it is still not clear to me how the distinction is made between OpenVSX and VS Marketplace for a package.

For example: https://marketplace.visualstudio.com/items?itemName=eamodio.gitlens  and https://open-vsx.org/extension/eamodio/gitlens would have the same name if I am reading the spec correctly? So how do I specify which ecosystem I am targeting with a report?

calebbrown:
Fixes #449.


another-rex (approved):
Nice, LGTM.

PR #463, opened 2025-11-27
feat(linter): use the ecosystem suffix as the repository when linting Packagist advisories

G-Rath:
This replaces the need for custom logic to handle Drupal packages while also showcasing how the new Packagist suffix works, following the same [logic that Composer does](https://getcomposer.org/doc/05-repositories.md#composer) to resolve a repository to its actual packages.

For now I have just added support for the [`metadata-url` field](https://getcomposer.org/doc/05-repositories.md#metadata-url-available-packages-and-available-package-patterns) as that is reasonably straightforward and used by the Drupal repositories.

Sadly, it seems that https://wpackagist.org and https://asset-packagist.org still use the Composer v1 [`provider-includes` and `providers-url`](https://getcomposer.org/doc/05-repositories.md#provider-includes-and-providers-url) fields which seems to be far more complex to support, but as we only have custom logic to handle the Drupal repository we're not losing functionality with this change

### DIFF
--- a/tools/osv-linter/internal/pkgchecker/packagist.go
+++ b/tools/osv-linter/internal/pkgchecker/packagist.go
@@ -6,10 +6,13 @@ import (
 	"net/http"
 	"net/url"
 	"strings"
+	"sync"
 
 	"github.com/ossf/osv-schema/linter/internal/faulttolerant"
 	"github.com/tidwall/gjson"
 )
+
+var packagistCache sync.Map
 
 func fetchPackagistPackagesInfo(repo string) ([]byte, error) {
 	resp, err := faulttolerant.Get(repo)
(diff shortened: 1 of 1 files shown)


tools/osv-linter/internal/pkgchecker/packagist.go
```diff
@@ -0,0 +1,89 @@
...
+	"net/http"
+	"net/url"
+	"strings"
+	"sync"
+
+	"github.com/ossf/osv-schema/linter/internal/faulttolerant"
+	"github.com/tidwall/gjson"
+)
+
+var cache sync.Map
```
another-rex: nit: This should just be for packagist, so let's rename this to packagistCache. 

In theory we can put all the ecosystem repo's in here, but I think it's probably clearer to have individual cache vars per ecosystem impl.

Issue #248, opened 2024-06-21
Proposal: add new severity[].source field


marco-silva0000:
Proposal to add a new optional string field on a severity entry that represents "who" scored or where that scoring came from.

Different entities score vulnerabilities differently and sometimes there are different sources that don't agree on scoring for the same vulnerability, this would allow the schema to support both instead of having to make a decision on which one is best.

oliverchang:
Hi, 

So sorry for missing this issue earlier! 

As OSV is a distributed database, where database owners publish their own vulnerability records, the implication is that all values in that record (including severity values) come from the database itself.

For example, if a GHSA advisory has a severity field, then the implication is that this severity comes from GitHub (or at least, GitHub endorses the severity if it came from somewhere else).

dodys:
Hi @oliverchang,

Just saw this issue and I think it would be nice to have a source field. Users/customers might be interested in seeing more than one rating in a single view. Therefore if Ubuntu wanted to have two different CVSS scores published in OSV, it would be unclear why they differ or who calculated/published it (for example one from NVD, one from Intel). It is more than just endorsing, it is also about users wanting this single view of all the different ratings.

litios:
I agree with @dodys here, it's not uncommon for distributions to provide their own CVSS and there is value in specifying who did the severity analysis. When looking at distro-specific CVEs, I don't know if the CVSS score listed there is specific to the distribution or it comes from the general CVE without having to consult both entries. 

Talking about CVE entries coming from NVD, their CVSS scores can be both their own ones as well as others, like 
[CVE-2024-49394](https://osv.dev/vulnerability/CVE-2024-49394). In this particular case, the analysis actually comes from Red Hat and there is no way to identify this from OSV data. 

Same happens with most NVD CVEs coming from a GitHub advisory, in which case the CVSS comes directly from GitHub.

andrewpollock:
Circling back to the point @oliverchang made in https://github.com/ossf/osv-schema/issues/248#issuecomment-2347745099 and attempting to restate it differently:

[...]

PR #510, opened 2026-02-03
Add a severity source field

conorfitch:
Hello! Please also see this related PR for osv.dev: https://github.com/google/osv.dev/pull/4729

* This PR adds an optional `severity[].source` field, previously discussed in this issue: https://github.com/ossf/osv-schema/issues/248

* There are several data sources in OSV.dev that aggregate severities from other providers, such as NVD or Ubuntu. This can lead to cases where the same severity is duplicated across aliased or upstream records, because the severity came from the same source. It can also result in cases where multiple severities are provided in a single record, but is unclear what the difference between them is.

* The addition of this field should help these issues to be solved, by allowing the source of a severity to be specified. For example, it can be ascertained through the field that one severity was issued by NVD, while the other was issued by the CNA.


conorfitch:
@jess-lowe That makes a lot of sense - to simply distinguish between CNA and NVD scores for example, would be excellent. 

That sounds similar to how NVD display the categories on their site - i.e. NVD, CNA, as well as CISA-ADP I think (not sure if they have any other categories). And in the case of a CVE, someone could be able to look at the "cna_assigner" from database_specific if it exists and they wish to do so.

Could it be that if the record doesn't provide a severity source, then it is to be assumed that the severity was produced by the record's provider? Or maybe there could be an explicit way of denoting that, for example, putting the record's id prefix as the severity source? 

It seems like a few databases copy the CVSS scores from NVD/CNA into their own OSV records currently, so I was wondering if those would ever end up being marked as such.

gregkh (commented):
> Signed-off-by: Conor Fitch <173908980+conorfitch@users.noreply.github.com>

Nit, I do not think this is not a valid email address for a signed-off-by line.

conorfitch:
Thanks @jess-lowe ! I like the sound of that - an enum with `NVD`, `CVE-CNA` and one to explicitly say it was provided by the record's database.

That is tricky to get the name right for the third option. The only suggestion I have to add to that list is `"source": "SELF"`.

In addition to `CVE-CNA`, would `CVE-ADP` be a reasonable option to add? Here's an example with scores from NVD, CNA and an ADP: https://nvd.nist.gov/vuln/detail/CVE-2025-9491. (Is it correct to have the `CVE-` prefix for it, or should it just be `ADP`?)

conorfitch:
@jess-lowe I've just updated this now if this looks alright 👍 - I'll update the other PR to OSV.dev shortly.

jess-lowe:
> @jess-lowe I've just updated this now if this looks alright 👍 - I'll update the other PR to OSV.dev shortly.

Awesome! Would we be able to also add these to the proto bindings?:

Edit https://github.com/ossf/osv-schema/blob/main/proto/vulnerability.proto then run https://github.com/ossf/osv-schema/blob/main/bindings/build.sh from the bindings directory :)

conorfitch:
Thanks that's added! One thing - `CVE-CNA` wasn't possible in the proto bindings (due to the dash). I've put the value as `CNA` for now, but would `CNA` or `CVE_CNA` be more preferable?

proto/vulnerability.proto
```diff
@@ -127,10 +127,19 @@ message Severity {
     CVSS_V4 = 3;
     Ubuntu = 4;
   }
+  // Source of the severity.
+  enum Source {
+    SOURCE_UNSPECIFIED = 0;
```
jess-lowe, line 132: I think because if it's unspecified, it's implicitly SELF, maybe we just set SELF as 0 and remove SOURCE_UNSPECIFIED?

conorfitch: Ah yes, that's a good point - is there any special meaning of "0" that this would have other implications for depending on how this is used?

Also would you say there's any semantic difference between specifying SELF vs omitting it, that should be preserved? e.g. omitting might not necessarily say that the db produced it (but it at least endorses it), whilst an explicit SELF states that the severity originates from that db - or do they just represent the same thing?

conorfitch: Hey @jess-lowe I'm just coming back to this now - I don't have much knowledge in this area but I was looking at the docs and found this section: https://protobuf.dev/best-practices/dos-donts/#unspecified-enum
Would it make sense to still keep `SOURCE_UNSPECIFIED` then based on that description?

another-rex: I can see it being fine both ways. Let's keep the UNSPECIFIED there then, in-case we ever want to change the default interpretation in the future.

FYI: Our team is traveling next week, so this will likely get merged in the week after.

jess-lowe (approved):
LGTM - thanks!

Issue #547, opened 2026-05-07
Tighten guidelines on acceptable IDs


jess-lowe:
This proposal formalizes the constraints on the id field within the OSV schema to improve automation and data integrity. 

### Character Set Constraints
To ensure maximum compatibility with CLI tools and URL routing, we propose a strict ASCII subset.

**Allowed Characters:** `a-z, A-Z, 0-9, :, -, _,` and `.`

**Regex:** `^[a-zA-Z0-9:_.-]+$`

**Rationale:** Eliminates the need for URL encoding and prevents shell-scripting errors caused by spaces or "creative" punctuation (e.g., ( ), !, or /).

### Semantic Neutrality (Opaque IDs)
We propose that IDs function solely as unique, permanent pointers rather than summaries.

**Guideline:** New IDs should be semantically meaningless. They must not describe the vulnerability type, package name, or severity. IDs should always start with the correctly scoped prefix, followed (ideally) with an `-`.

**Rationale:** Metadata changes. If an ID is OSV-LOG4J-CRITICAL and the severity is later downgraded or the package name is found to be different, the ID becomes a source of misinformation. Opaque IDs (e.g., OSV-2026-X7Y2) stay accurate forever.

PR #557, opened 2026-06-03
feat(osv-linter): add REC:005 id character-set check

arpitjain099:
## Summary

This adds a new osv-linter record check, `REC:005` (`record-id-valid`), that validates the OSV `id` field against the character set proposed in #547: `^[a-zA-Z0-9:_.-]+$`. An `id` containing any character outside that ASCII subset (for example a space or a `/`) now produces a finding.

The rationale follows the issue: constraining ids to this subset avoids the need for URL encoding and prevents shell-scripting errors caused by spaces or other punctuation in downstream tooling.

## Scope

#547 is a two-part proposal. I deliberately scoped this PR to only the first, mechanical part (the character-set constraint), because it is objectively checkable with a regex. I did not implement the second part (the "semantic neutrality / opaque IDs" guideline that new ids should be semantically meaningless), since that is a human-judgement guideline rather than something a linter can reliably enforce, and it reads more like documentation guidance than a deterministic check.

## Implementation

The check mirrors the existing `REC:00x` checks exactly:

- a package-level `validIDPattern` compiled regex in `internal/checks/record.go`
- a `CheckRecordHasValidID` `CheckDef` (code `REC:005`, name `record-id-valid`)
- an `IDCheck` function that reads `id` via gjson and emits a `CheckError` on a non-conforming value (records without an `id` are left to the schema check and are not flagged here)
- registration in the `ALL` and `offline` collections (it has no remote data dependency, so it is an offline check; I left it out of `fatal`, consistent with `REC:004` also not being in that curated subset)

It uses only the standard library (`regexp`, `fmt`); no new dependency is added.

## Testing

`TestIDCheck` is added to `record_test.go` following the existing table-test style: valid ids pass (a normal `CVE-` id and one exercising the permitted `: . - _` punctuation), a missing `id` is a no-op, and ids containing a space or a slash fail with the expected message.

- `go test ./internal/checks/...` passes (full module `go test ./...` also green)
- `go vet ./internal/checks/...` is clean
- the new files are `gofmt` clean
- `go run ./cmd/osv record check --checks list` shows `REC:005: (record-id-valid): id field uses only permitted characters`

## Note

I opened this as a focused first step on #547. As far as I can tell no maintainer has weighed in on the issue yet, so I am happy to adjust the naming, the message wording, which collections it belongs to, or whether the character set should match here exactly, based on your preference. Happy to also pick up the semantic-neutrality half separately if you decide it should live as a (likely documentation-only) guideline.

Addresses #547.

### DIFF
--- a/tools/osv-linter/internal/checks/checks.go
+++ b/tools/osv-linter/internal/checks/checks.go
@@ -104,6 +104,7 @@ var Collections = []CheckCollection{
 			CheckRecordHasValidAliases,
 			CheckRecordHasValidUpstream,
 			CheckRecordHasValidRelated,
+			CheckRecordHasValidID,
 			// Range checks
 			CheckRangeHasIntroducedEvent,
 			CheckRangeIsDistinct,
@@ -123,6 +124,7 @@ var Collections = []CheckCollection{
 			CheckRecordHasValidAliases,
 			CheckRecordHasValidUpstream,
 			CheckRecordHasValidRelated,
+			CheckRecordHasValidID,
 			CheckRangeHasIntroducedEvent,
 			CheckRangeIsDistinct,
 			CheckPackagePurlValid,
--- a/tools/osv-linter/internal/checks/record.go
+++ b/tools/osv-linter/internal/checks/record.go
@@ -1,8 +1,16 @@
 package checks
 
 import (
+	"fmt"
+	"regexp"
+
 	"github.com/tidwall/gjson"
 )
+
+// validIDPattern is the set of characters permitted in an OSV record id.
+// Restricting ids to this ASCII subset avoids the need for URL encoding and
+// prevents shell-scripting errors caused by spaces or other punctuation.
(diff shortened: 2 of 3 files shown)


jess-lowe (approved):
LGTM. Thanks! 

note: validated the regex on all current files as of 9 Jul 26 in osv.dev's all.zip

arpitjain099:
@jess-lowe thanks again for validating the regex against the osv.dev corpus. Approved since 9 July and still open, does it need anything more from me?

jess-lowe:
Apologies, @arpitjain099! For some reason the CI has been stuck not running the checks so I need to figure out how to make it run so I can press merge. Would you be able to send an empty commit to try retrigger the checks?

arpitjain099:
Done, pushed an empty commit. Checks have run and are green.

jess-lowe:
Thank you!

PR #560, opened 2026-06-10
feat: support wildcard package name '*' in schema and linter

another-rex:
This PR introduces support for a special wildcard package name `*` to the OSV schema and the `osv-linter` validation tool.
## Changes
- Update the schema.md to document the wildcard package name `*` which allows an advisory to affect all packages within a specified ecosystem (e.g. to denote EOL).
- Also updated the linter and jsonschema

This fixes #519 by providing a mechanism to represent End-of-Life (EOL) for an entire ecosystem without having to list every individual package. By using a wildcard package name `*`, data providers can publish a single advisory for the EOL ecosystem version (e.g. `Ubuntu:18.04`). Any query for a package within that ecosystem will then match this advisory, alerting users that the ecosystem itself is no longer supported, so the results might not be complete / up to date.

Definitely looking for feedback on this approach!

### DIFF
--- a/docs/schema.md
+++ b/docs/schema.md
@@ -887,6 +887,11 @@ strings in the table below. The `name` field is a string identifying the library
 within its ecosystem. The two fields must both be present, because the
 `ecosystem` serves to define the interpretation of the `name`.
 
+Use the wildcard package name `*` to indicate that an advisory affects all
+packages within the specified `ecosystem`. For example, this is useful when an
+ecosystem version reaches end-of-life (EOL) and receives no further security
+support.
+
 The `purl` field is a string following the
  [Package URL specification](https://github.com/package-url/purl-spec) that
 identifies the package, without the `@version` component.
--- a/tools/osv-linter/internal/checks/packages.go
+++ b/tools/osv-linter/internal/checks/packages.go
@@ -44,6 +44,9 @@ func PackageExists(json *gjson.Result, config *Config) (findings []CheckError) {
 			return true // keep iterating (over affected entries)
 		}
 		pkg := value.Get("package.name").String()
+		if pkg == "*" {
+			return true // keep iterating, wildcard package matches everything and doesn't need to exist
+		}
 
 		// Avoid unnecessary network traffic for repeat packages.
 		if _, ok := knownExistent[Package{Ecosystem: ecosystem, Name: pkg}]; ok {
@@ -92,6 +95,9 @@ func PackageVersionsExist(json *gjson.Result, config *Config) (findings []CheckE
 			return true // keep iterating (over affected entries)
 		}
 		pkg := value.Get("package.name").String()
+		if pkg == "*" {
+			return true // keep iterating, wildcard package matches everything and doesn't need to exist
+		}
 		versionsToCheck := []string{}
 		// Examine versions in ranges.
 		maybeRanges := value.Get("ranges")
--- a/tools/osv-linter/internal/checks/packages_test.go
+++ b/tools/osv-linter/internal/checks/packages_test.go
@@ -27,6 +27,14 @@ func TestPackageExists(t *testing.T) {
 			},
 			wantFindings: []CheckError{{Code: "", Message: "package \"123bla\" not found in \"PyPI\""}},
 		},
+		{
+			name: "Wildcard package name in PyPI",
+			args: args{
+				json:   LoadTestData("../../testdata/wildcard-package.json"),
+				config: &Config{},
+			},
+			wantFindings: nil,
+		},
 	}
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
@@ -62,6 +70,14 @@ func TestPackageVersionsExists(t *testing.T) {
 				json:   LoadTestData("../../testdata/GHSA-9v2f-6vcg-3hgv.json"),
 				config: &Config{Ecosystems: []string{"npm"}},
 			},
+		},
+		{
+			name: "Wildcard package versions check",
+			args: args{
+				json:   LoadTestData("../../testdata/wildcard-package.json"),
+				config: &Config{},
+			},
+			wantFindings: nil,
 		},
 	}
 	for _, tt := range tests {
--- a/tools/osv-linter/testdata/wildcard-package.json
+++ b/tools/osv-linter/testdata/wildcard-package.json
@@ -1,0 +1,19 @@
+{
+  "modified": "2026-06-10T00:00:00Z",
+  "published": "2026-06-10T00:00:00Z",
+  "schema_version": "1.5.0",
+  "id": "CVE-2026-99999",
+  "summary": "Ecosystem wildcard test",
+  "details": "This is a test case for a wildcard package name.",
+  "affected": [
+    {
+      "package": {
+        "ecosystem": "PyPI",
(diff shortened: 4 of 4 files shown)


dodys:
@jasnow @DmitriyLewen  fyi

DmitriyLewen:
Hello all
@dodys Thanks for ping.

Correct me if I'm wrong:

The idea behind this change is to mark a release as EOL — i.e. a package entry with `name: "*"` means the advisory applies to the whole ecosystem, and here it's used to signal that the release has reached EOL.

This helps with the problem of records simply "disappearing" after EOL, but it doesn't really help scanners.

From a scanner's point of view, we now at least know that the release is EOL, rather than it being missing from the data for some unknown (for scanner) reason. (Worth noting: Ubuntu's OSV records only list *affected* packages/releases — there is no "not-affected" representation, so today a missing release is ambiguous: it could be not-affected, or just no longer loaded because it's EOL. The wildcard removes that ambiguity.)

But we still don't know the affected packages: we don't know whether all known vulnerabilities were fixed before EOL, or whether something was found after EOL that will never be fixed, and so on.

another-rex:
Yes, that's mostly correct.

> we don't know whether all known vulnerabilities were fixed before EOL

If a known vulnerability is fixed before EOL, the advisory for that vulnerability should contain a fixed version that's released within that ecosystem.

> or whether something was found after EOL that will never be fixed, 

This it doesn't show this information, but from what I understand, that's what EOL is describing - this ecosystem/release is no longer maintained, so no one is processing or publishing any vulnerability reports, and you should move off this release, as packages used within will be a blind spot for any scanners.

DmitriyLewen:
Am I understanding correctly that you are proposing to always keep a record for a specific release?

- Keep the correct package name for an Release entry with:
	- a fixed version before EOL;
	- Confirmed vuln, but there is no fix and the release going EOL;
	- if the bug was found after EOL (for example, the same component version is installed in both a newer and an older OS version; a vulnerability was found in the newer OS and the component in the older OS was marked as vulnerable — if I understand https://ubuntu.com/security/CVE-2024-0564 (`Ignored end of standard support, was needed`) correctly, this is that case; @dodys please correct me if I'm wrong).
- Use the wildcard for:
	- unconfirmed vulnerabilities.

If that's the case, then this is a solution for scanners, but this PR isn't really related to the solution, because these changes only extend the information (EOL, or the package is not vulnerable).

If you do think that the past records (with fixed versions, etc.) should be replaced with a wildcard, then the wildcard solution will not help scanners. At the moment of EOL (a day or two after the records are replaced), users will stop receiving advisories for that OS.
Upgrading the OS is a long and complex process for many (the "bigger" the user, the harder it is).
But until they upgrade, users need to know that their OS is vulnerable (e.g. if they are not yet on the latest version). This serves as an additional motivator for users to upgrade the OS.

another-rex:
I'm not quite sure I understand your comment. This proposal does not apply to existing records. This is the rough timeline I'm imagining,

1. **Vulnerability Found:** A vulnerability for `pkg: foo` in ecosystem `abc:12` is found.
2. **Fix Developed:** A fix for `pkg: foo` in `abc:12` is made.
3. **Advisory Published:** An advisory for `pkg: foo` in `abc:12` is published with the fix version.
4. **Ecosystem Goes EOL:** `abc:12` goes End-of-Life (existing advisories still exist).
5. **EOL Blanket Advisory Published:** An EOL advisory is published for `abc:12`, which matches all packages in the ecosystem. (*What this proposal enables*)
6. **New Vulnerability Found:** A new vulnerability is later found for `pkg: foo` in ecosystem `abc:12`.
7. **No New Advisories Published:** No new specific advisories are published for this vulnerability because the ecosystem is EOL.

Currently the same flow is happening, but step 5 is missing.

> Upgrading the OS is a long and complex process for many (the "bigger" the user, the harder it is).

Could adding a date help with this? An advisory that's time bound, to let the user know that the EOL is coming up, and they should look at starting to migrate off e.g. 6 months earlier.

DmitriyLewen:
Since the problem started with the removal of records from Ubuntu, a lot of my reasoning is based specifically on that experience.
Thanks a lot for the detailed example.

> Currently the same flow is happening, but step 5 is missing.

Now I understand your logic, and this works for us.
@dodys, does this approach work for you?

> Could adding a date help with this?

A date can definitely be useful for scanners — but I'd argue it belongs in a separate, machine-readable source maintained by the vendor with the official EOL dates.

The key reason is that these dates aren't static: some vendors revise them over time (Amazon Linux, for example, has pushed its EOL dates back more than once). Keeping them in one authoritative, machine-readable source means scanners always read the current value, and OSV stays focused on the vulnerability data — while the wildcard advisory still carries the present-tense "this release is EOL now" signal.

dodys:
I think my biggest question is, where would this blank advisory just for the EOL would live? 
Should we create under our repo something like: `osv/eol/UBUNTU-EOL-25.10.JSON` ? And would we need a new ID for it (like the example before)? 
For the date, we could add to the same blank advisory, under ecosystem_specific.

another-rex:
> I think my biggest question is, where would this blank advisory just for the EOL would live? 
> Should we create under our repo something like: `osv/eol/UBUNTU-EOL-25.10.JSON` ? And would we need a new ID for it (like the example before)? 
> For the date, we could add to the same blank advisory, under ecosystem_specific.

Yes it would need to be a new ID, as for where it will live, from the scanner / ingestors perspective this shouldn't matter too much.

jess-lowe (approved):

jasnow:
@another-rex - Can you show an example for my Rubygems ecosystem? It does not have to be a real example, just one to see what level you are EOF'ing. Thanks.

another-rex:
Sure! Though some clarification first:

This wildcard change specifically doesn't really affect the Rubygems ecosystem at all (except in the unlikely case that all RubyGems gets deprecated), as it's mostly focused on ecosystems that have :<version> suffixes.

To mark packages as deprecated/EOL in RubyGems, you can use the existing method of publishing a record with an `introduced: 0` and no `fixed` version, and the package name being the package that's EOL. Here's an example from the crates.io ecosystem: https://osv.dev/vulnerability/RUSTSEC-2026-0214

Or as an example RubyGems record:

```
{
  "id": "GHSA-1234-5678",
  "summary": "xyz is unmaintained",
  "details": "The xyz gem is unmaintained, and all versions are affected as no more security advisories will be published for this package.",
  "modified": "2026-07-23T07:30:03.563653170Z",
  "published": "2026-07-22T12:00:00Z",
  "references": [
...
  ],
  "affected": [
    {
      "package": {
        "name": "xyz",
        "ecosystem": "RubyGems",
      },
      "ranges": [
        {
          "type": "SEMVER",
          "events": [
            {
              "introduced": "0"
            }
          ]
        }
      ]
    }
  ],
  "schema_version": "1.9.0"
}
```

jasnow:
@another-rex - Thanks for the example.

FYI: My guess is that [rubysec.com](https://github.com/rubysec/ruby-advisory-db) will inherit this information from repo or global GHSA, CVE, or OSVDB (do not expect new advisories from there) databases. where
  * repo GHSA is https://github.com/PROJECT/REPO/security/advisories/GHSA-abcd-efgh-ijjl
  * global GHSA is https://github.com/advisories/....

PR #570, opened 2026-06-29
Add vcpkg ecosystem

### LINKED ISSUE #561: Add vcpkg ecosystem
I'd like to add `vcpkg` as a supported OSV ecosystem so we can publish security advisories for [vcpkg](https://github.com/microsoft/vcpkg) packages in OSV format.
 
## Background

vcpkg is Microsoft's open source C/C++ package manager. Ports come from a git registry (the default is [`microsoft/vcpkg`](https://github.com/microsoft/vcpkg)), and a package version is the upstream version plus an integer `port-version` that bumps when the port file changes. There are no distro-style releases, so the ecosystem string has no release suffix.

## Proposed schema

- Ecosystem name: `vcpkg`
- Format: `vcpkg` (no release suffix; the registry/baseline is captured by the purl qualifiers below)
- Package naming: `name` is the vcpkg port name (e.g. `bzip2`, `ffmpeg`).
- Versions: upstream version plus `port-version`. The exact ordering still needs pinning down, since vcpkg ports use several version schemes (semver, date, plain string). I'll write that up before the PR.
- PURL: `pkg:vcpkg/<name>@<version>?port_version=<n>` (with optional `repository_url` / `repository_revision` for non-default registries)
- ID prefix: `VCPKG` (proposed, e.g. `VCPKG-0001`)
 
## Dependency

The `vcpkg` PURL type isn't registered yet. It's in progress at https://github.com/package-url/purl-spec/pull/562 (continues https://github.com/package-url/purl-spec/pull/245, closes https://github.com/package-url/purl-spec/issues/217), and I'm helping land it. OSV.dev maps every ecosystem to a purl type in `purl_helpers.py`, so that PR is effectively a prerequisite and should merge first.
 
## References

- vcpkg: https://github.com/microsoft/vcpkg
- PURL type PR: https://github.com/package-url/purl-spec/pull/885
- port-version reference: https://learn.microsoft.com/vcpkg/reference/vcpkg-json#port-version

JamieMagee:
Adds `vcpkg` as an OSV ecosystem and registers the `VCPKG` advisory prefix.

The purl type it needs ([purl-spec#885](https://github.com/package-url/purl-spec/pull/885)) is already merged. So this declares the ecosystem in `ecosystems.json`, then regenerates the derived files with `scripts/update-ecosystems-lists.py`. Versions are the upstream port version, with `port-version` carried in the purl `port_version` qualifier. Source URLs for the home database are TBD for now, matching how Azure Linux landed.

Closes #561.


jess-lowe:
This LGTM, but I'll wait until @another-rex is back for a second pair of eyes.

docs/schema.md
```diff
@@ -634,6 +634,17 @@ The defined database prefixes and their "home" databases are:
         </ul>
       </td>
     </tr>
+    <tr>
+      <td><code>VCPKG</code></td>
+      <td><a href="https://github.com/microsoft/vcpkg">vcpkg Security Advisory</a></td>
+      <td>
+        <ul>
+          <li>How to contribute: TBD</li>
```
another-rex, line 642: What's the best way for people to contact you if they find issues/questions with the data?

another-rex (commented):
LGTM, would be good to get some form of contact details in case there are issues with the data though.

another-rex (approved):
Actually if this is published like AzureLinux records via a github repo where people can just post issues, then this is fine!

JamieMagee:
@another-rex thanks for the approval! Yes, we plan to use GitHub to host the advisories.

PR #573, opened 2026-07-09
feat: add include-withdrawn flag to linter and skip withdrawn by default

rjerrems:


tools/osv-linter/internal/linter_test.go
```diff
@@ -0,0 +1,72 @@
...
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			contentBytes, err := os.ReadFile(tt.filename)
+			if err != nil {
+				t.Fatalf("failed to read test file %q: %v", tt.filename, err)
+			}
+
+			// Inject a duplicate entry into "related" to ensure a finding is triggered when run
+			badRelated := []byte(`"related": ["CVE-2021-44832", "CVE-2021-44832"],`)
+			contentBytes = bytes.Replace(contentBytes, []byte(`"related": [`), badRelated, 1)
```
another-rex: I think this replace results in an invalid JSON? You might need to remove the last `],` in the badRelated var.

rjerrems: Good catch, I have changed it to properly parse the JSON and replace that key with a known bad value.

another-rex (approved):

PR #588, opened 2026-08-03
docs: clarify guidance on multiple ecosystems and GIT entries

jess-lowe:
This PR updates the schema documentation to provide clear guidelines on how to structure `affected` fields when a vulnerability applies to both a strict/"real" ecosystem (like `Go` or `npm`) and generic Git repositories.

- Explains how to depict multiple ecosystems in the same record.
- Recommends using distinct entries in the `affected` array for `GIT` and strict ecosystems to ensure tooling compatibility.
- Clarifies that `GIT` records get their `versions` enumerated from the attached repository tags, and how OSV infrastructure automates this.

Mixing Git commit hashes and semantic versions in the same `affected` entry can cause mixed enumerated versions in the versions array, that can confuse both publishers and consumers. This documentation guides publishers to separate them for better compatibility.

### DIFF
--- a/docs/schema.md
+++ b/docs/schema.md
@@ -1005,6 +1005,23 @@ It is permitted for a database name (the DB prefix in the `id` field) and an
 ecosystem name to be the same, provided they have the same owner who can make
 decisions about the meaning of the `ecosystem_specific` field (see below).
 
+### Depicting multiple ecosystems in the same record
+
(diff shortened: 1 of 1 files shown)


G-Rath (approved):
this looks good, though I wonder if something should be changed with the `affected[].versions field` a few headers down?

Not sure exactly what the change is, it just feels a little odd for that to have no changes...

jess-lowe:
> Not sure exactly what the change is, it just feels a little odd for that to have no changes...

Updated it to point at the other paragraph and be a little more specific

another-rex (approved):

PR #611, opened 2026-09-15
linter: check last_affected in RNG:002

deep-sengupta:
## Summary

Fix `RNG:002` so that `last_affected` events are checked when determining whether an introduced and ending event overlap.

## Problem

`RangeIsDistinct` documents that it checks `introduced` against `fixed` or `last_affected`, but the implementation only collected `fixed` events.

As a result, a range containing:

```json
{
  "introduced": "1.0.0"
},
{
  "last_affected": "1.0.0"
}
```

was not reported by `RNG:002`.

## Fix

Add `last_affected` values to the range ending events checked by `RangeIsDistinct`.

A regression test was added covering:

* identical `introduced` and `last_affected` values
* different `introduced` and `last_affected` values
* existing `introduced` and `fixed` behavior

## Validation

* `go test ./internal/checks` ✅
* `go test ./...` ✅
* `git diff --check` ✅
* Verified the CLI reports `RNG:002` for identical `introduced`/`last_affected` values.


jess-lowe:
Technically speaking, `introduced` and `last_affected` are allowed to be identical / overlap (as mentioned on [line 75](https://github.com/ossf/osv-schema/pull/611/changes#diff-68643b97225aaabac165ddb0567c88e66e6fb8ac1c0677f6cda6213b35212b6dL75)), as it is currently the best way of presenting that only that specific version is affected. 

`last_affected` and `fixed` definitely shouldn't overlap, but I believe this should be being checked in a different linter rule.

PR #615, opened 2026-09-22
docs: document the Echo ecosystem's language suffixes

orizerah:
## Summary

Updates the `Echo` ecosystem definition to document the `:<ECOSYSTEM>` suffixes, as requested by @another-rex in [google/osv.dev#6019](https://github.com/google/osv.dev/issues/6019):

> I also notice that on the schema the Echo ecosystem doesn't actually mention the `:PyPI` or other ecosystems can be appended. Can you please update the schema with the updated definition?

## Background

Echo is a Debian-derived distribution that also publishes secured builds of packages from other ecosystems. Which one an advisory refers to is given by a suffix on the ecosystem string:

| Ecosystem | Packages | Version ordering |
|---|---|---|
| `Echo` | Echo OS packages | Debian |
| `Echo:PyPI` | Python | PEP 440 |
| `Echo:Maven` | Maven | Maven |
| `Echo:npm` | npm | SemVer |

The existing definition described only the unsuffixed form, so neither the suffixes nor their canonical spelling were documented. The spelling in particular has already caused a real integration break — the exported data used `Echo:PyPi` rather than `Echo:PyPI`, which broke downstream scanner matching until it was re-imported.

## Changes

Updates the `Echo` entry in `ecosystems.json` and regenerates the list in `docs/schema.md` with `scripts/update-ecosystems-lists.py`, spelling out:

- that the ecosystem string may carry a `:<ECOSYSTEM>` suffix, and what `name` means when it does
- the supported values, in their canonical spelling
- which version ordering applies with and without a suffix
- the `+echo.<N>` build suffix that Echo's secured builds carry (e.g. `1.2.3+echo.1`)

No change to `validation/schema.json` — `ecosystemWithSuffix` already permits `(:.+)?`, and the linter (`pkgchecker`) already handles Echo independently of its suffix. `scripts/validate-schema-table.py` passes.


another-rex (approved):
Nice, thanks!

